PR #7, opened 2016-04-03
Fixed error message condition for /premium

NorbiPeti:


games647:
Oh, you're right. Thank you.

PR #8, opened 2016-04-03
Allow automatic logins if autoRegister is on

NorbiPeti:


NorbiPeti:
Also, I don't see how can this be insecure. Premium users are supposed to have different UUIDs than cracked players, so how'd they gain access on name change? Or what other problems are there?

games647:
Premium players only have different UUIDs if you activated premium UUID in the config. Otherwise they would have an offline uuid. Moreover many plugins  (including auth plugins) are still not UUID compatible and uses the player name as a unique id.

I think this config option is misleading again. I actually is meant to be "auto assume premium". On a login request we perform a check if this name is associated to a premium account. If the user is not in our premium list and is not registered yet (-> no cracked player has this account on this server), we assume it's a premium player and request a premium authentication. 

In the case cracked players who just uses a premium username wants to join the server, they will disconnect with the message "bad login". There is no way to change this message or remove this behavior, because it's connection between the client and the mojang servers.

games647:
Therefore we check if the account exists (->could be a cracked player) and if this is not the case enable encryption (premium login).

NorbiPeti:
Our problem is that we are restarting the server everyday, therefore the
plugin doesn't seem to have an use for us as it is now. Premium players
still have to log in, then type /premium incase they need to relog. What I
had in mind was an option to avoid having to use /login at all for premium
players (we could take care of not allowing users with a registered name).
Or is that how it works already, it was just that I registered my name
before I enabled the autoregister option?

games647:
>Premium players
still have to log in, then type /premium incase they need to relog. 

That's why the plugin is still in beta. I want to change this behaviour for the next version. There are two option. Either save and read a file for the the list of activated premium players or implement a database. The last option needs more time to implement, but it makes it possible to let players join even if the mojang session servers are down. It would also provide better compatibility with the BungeeCord version of this plugin, because BungeeCord can then contact instead of communicating with through the plugin channels. 

>Or is that how it works already, it was just that I registered my name
before I enabled the autoregister option?

This isRegistered check exist there to make this account isn't owned by a cracked account, because a premium player would auto login into it. So you can understand it as check to prevent premium players from stealing the account of cracked ones. 

But yes these all checks are only enabled if the player has **premium username** and is **not a bungeecord user**. (it's not implemented for BungeeCord yet. See the planning above)

Source: 
Reference/Call:  [StartPacketListener](https://github.com/games647/FastLogin/blob/master/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/StartPacketListener.java#L89)
Declaration: [MojangApiConnector](https://github.com/games647/FastLogin/blob/master/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/MojangApiConnector.java#L41)

NorbiPeti:
Okay, so auto-registered users get logged in automatically. Thanks, I did not know that. :P Can you tell me how that works (just for my curiousity), since it auto-logins me even if I restart the server.

games647:
>Though I think it should be an option to auto-login as well, since the server owners could simply warn cracked players to use a name that isn't taken already, so there is no chance of their playerdata to be shared with a premium account.

I think it will be unnecessary after the database is implemented, because all new players will see this behaviour and the existing players could type /premium instead. 

>Actually, I think storing if a user is premium is the best, while allowing premium users to log in and register automatically, and if a cracked player tries to log in with that name, it wouldn't allow it. 

This is what I intend to do with the option autoRegister.

>And same for the opposite way, if a cracked account name is bought by a player it wouldn't let them join with that name until it is renamed or something like that.

Currently if a premium user joins the server and a cracked account already exists, the plugin still lets them join the server, but without auto login. So it asks for the password, because there is no way to determine (except the IP) if it's not the same person. There could be the case that this cracked player bought the account.

PR #34, opened 2016-07-08
Add option to protect Premium User Name

mics8128:
see title
i only test with spigot
didn't test bungee.

for now, want to force cracked player join server with premium user name have to do below

>premium username
>cracked username

and join server..


bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/LoginSkinApplyListener.java
```diff
@@ -20,7 +20,7 @@ public LoginSkinApplyListener(FastLoginBukkit plugin) {
     }
 
     @EventHandler(priority = EventPriority.LOW)
-    public void onPlayerLogin(PlayerLoginEvent loginEvent) {
+    public void onPlayerLogin(PlayerJoinEvent loginEvent) {
```
games647: Why do you changed this?

games647:
Isn't this the same as autoRegister where it checks if the username is premium and if the username isn't registered by a cracked player. Then it thinks it's a premium player and save it as such after a successful response.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/LoginSkinApplyListener.java
mics8128: on my test server get null on player.getAddress()
that why i change this.

games647: So player was null?

How can that be null? Do had any plugins besides ProtocolLib on your server? I tested it recently and it worked.

mics8128:
no, not totally same,
because if any new premium player want join server,
but his name already used by other cracked player.
that new premium player can't login anymore.

that is not good feeling for new premium player.
that why i want to add this option to protect premium accounts.

let cracked player only can use non-premium username.

sorry for my poor English.

games647:
But autoRegister does the same if a cracked player doesn't exist on the server. 

Plus it automatically registers the player, so there is no need to do /register pass.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/LoginSkinApplyListener.java
mics8128: http://pastebin.com/jYbgFW17
this is my server log 
fastlogin is use your newest source code to build

not player is null
player.getAddress() is null (i split code to find out it)

games647: Yeah I saw it by attaching the debugger. Weird behaviour. 

I'll fix that.

games647:
As far as I know the code does exactly the same except checking if the account exists.

games647:
Sorry for being so mean.

So what you want is letting **only premium players choosing a premium name**, but unlike autoRegister **not automatically register** the player?

mics8128:
Sorry... my mistake, maybe my first pull comment not good, let you didn't know what i want.

below is why i want add this option:

server always have new player join,
but if new premium user know his name is used by others cracked player
he maybe leave. right?

that is why i want to make sure all cracked player can't use premium username

games647:
Then the same with autoRegister. If this is activated, **new** cracked player cannot choose a premium username.

mics8128:
hmm, yes
but cracked player change his name is easier than premium user.

and I have another problem:
now i have online-mode server with many premium players
it didn't install any auth plugin
now i want turn server from online-mode to offline-mode.

in this case, 
if fastlogin don't have protect premium username option
cracked player have chance to steal premium player account.
that why this option is important for me now. ( this is real case :P )

----------------------------------------------------------------------
if you think this option is not necessary
or you have better way to solve this problem
just close this pull request. :)

games647:
>but cracked player change his name is easier than premium user.

Yes I know, so this option autoRegister exists.

>if fastlogin don't have protect premium username option
cracked player have chance to steal premium player account.
that why this option is important for me now. ( this is real case :P )

But then you could also make use of autoRegister.

Every premium player will be automatically registered. Cracked players using a premium username can**not**  join the server at all.

Sorry if this is confusing for you.

mics8128:
Oh.... 
you are right... 
sorry for my misunderstand....

because i saw it call auth isRegister function,
so i think that only check auth is registed or not....

i delete authme and fastlogin database
and set autoRegister to True
and use cracked client whit my premium name,
can't join game..

but i don't know why authme API isRegister will return true when player not exist....

sorry for take you long time :P

games647:
>but i don't know why authme API isRegister will return true when player not exist....

Maybe because it's cached. Try to restart the server after deleting the database.

mics8128:
Oh.. i know the reason!

because `plugin.getConfig().getBoolean("nameChangeCheck")` in 
`if (plugin.getConfig().getBoolean("nameChangeCheck") || (plugin.getConfig().getBoolean("autoRegister")
                        && plugin.getAuthPlugin().isRegistered(username))) {
                    premiumUUID = plugin.getCore().getMojangApiConnector().getPremiumUUID(username);
                }`
 will let premiumUUID is NOT null.

and `premiumUUID != null && plugin.getConfig().getBoolean("autoRegister")` is always true when any player with premium name...

sorry for take you long time again :/
my fault, i didn't read config detail.

games647:
Thanks for pointing that out. You're right that is wrong

PR #41, opened 2016-07-19
Whitelist for cracked player

ac9607:
Hello
I was migrated from Auto-in
Auto-in has a feature:
Whitelist for cracked player, and premium player can login bypass whitelist
Can you add this feature?


games647:
A pull request is for contributions, please use issues instead.

PR #144, opened 2017-04-18
Fixed AuthHook

DevLeoko:
The setServerStarted()-Method is now also called if an extern AuthHook
hooks into FastLogin via the API


bukkit/src/main/java/com/github/games647/fastlogin/bukkit/tasks/DelayedAuthHook.java
```diff
@@ -35,10 +29,14 @@ public void run() {
                     + "and bungeecord is deactivated. "
                     + "Either one or both of the checks have to pass in order to use this plugin");
         }
+
+        if(hookFound && !plugin.isServerFullyStarted())
```
games647: This could be simplified to if (hookFound)

DevLeoko: Oh ya you are right, just changed it ;)

PR #249, opened 2019-05-06
Update LoginSecurity repository

lenis0012:
LoginSecurity is now on CodeMC.
I triggered a build for v2.1.9, and it will be the primary CI/repo from v3.0 onwards.


lenis0012:
Im sorry for making 2 commits, unfortunately I cannot use the git cli right now so I cannot fix it.

PR #287, opened 2019-12-08
Update ProtocolSupport version to support latest ProtocolSupport development builds

ishland:


ishland:
Tested on git-Paper-234 (MC: 1.14.4)
with ProtocolSupport 03cd0161f6277eb41ce8311741fb2564df88ce41 build 456
ViaVersion 2.1.4-1.15-pre6 build 296
ProtocolLib 4.5.0 build 429

games647:
Were there any incompatible changes? Normally it shouldn't be necessary to upgrade the dependency version.

ishland:
Yep, but if you don't update the dependency version, it is incompatible with newer ProtocolSupport.
I will test this on lower versions later.

games647:
This would mean that were some binary incompatible changes, because nothing of the API usage has changed. This unlikely, but can happen.

PR #360, opened 2020-06-07
Add a workaround fix for CatServer / Magma servers

NewbieOrange:

### Summary of your change
This PR contains a workaround fix for ProtocolLib wrongly identifying PacketType under bukkit-forge hybrid servers like CatServer and Magma. The issue caused FastLogin throws errors when client sends PacketLoginInStart packet, but incorrectly handled as PacketLoginEncryptionBegin.

### Related issue
None?

### DIFF
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/ProtocolLibListener.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/ProtocolLibListener.java
@@ -53,8 +53,7 @@ public class ProtocolLibListener extends PacketAdapter {
         }
 
         Player sender = packetEvent.getPlayer();
-        PacketType packetType = packetEvent.getPacketType();
-        if (packetType == START) {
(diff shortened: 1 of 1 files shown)


NewbieOrange:
Here is the exception message under CatServer 1.12.2: https://pastebin.com/sYJGCyup

games647:
Doesn't it make more sense to fix it upstream directly -> in ProtocollLib? Not only FastLogin could have issues because of it.

NewbieOrange:
It's difficult to fix this in upstream because of  the difference in obfuscation of forge, and the maintainers seem to have no plan to fix it too. I think it would be nice if fastlogin can have this workaround until ProtocolLib fixes this or 1.12 forge servers are no longer popular.

NewbieOrange:
This is probably related to dmulloy2/ProtocolLib#825.

games647:
I subscribed to the linked issue. I could implement this workaround, but I think checking for the gameprofiles could be too weak. Even though it's unlikely that something changes there it could easily break between versions.

NewbieOrange:
Fixed in upstream.

games647:
Thanks.

PR #367, opened 2020-06-25
Implement FastLoginPremiumToggleEvent

sgdc3:

### Summary of your change
Implemented a FastLoginPremiumToggleEvent event, useful to notify other plugins about premium status changes

### Related issue
None


games647:
Doesn't it make sense to also include the target state what the player wants change to.

sgdc3:
@games647 hmm, i think it makes more sense to include the previous state, as the new one can be obtained by the stored profile

games647:
Fair point.

Issue #417, opened 2020-11-07
error encryption


cicatrice74:
i have this error on latest dev of paper-spigot 1.16.4 and latest dev of protocollib and latest dev of fastlogin:

[22:33:39 ERROR]: [FastLogin] Couldn't enable encryption
java.lang.IllegalArgumentException: Unable to find a in net.minecraft.server.v1_16_R3.NetworkManager
at com.comphenix.protocol.reflect.FuzzyReflection.getMethodByParameters(FuzzyReflection.java:261) ~[?:?]
at com.github.games647.fastlogin.bukkit.listener.protocollib.VerifyResponseTask.enableEncryption(VerifyResponseTask.java:180) ~[?:?]
at com.github.games647.fastlogin.bukkit.listener.protocollib.VerifyResponseTask.verifyResponse(VerifyResponseTask.java:87) ~[?:?]
at com.github.games647.fastlogin.bukkit.listener.protocollib.VerifyResponseTask.run(VerifyResponseTask.java:63) ~[?:?]
at java.util.concurrent.CompletableFuture$AsyncRun.run(Unknown Source) ~[?:1.8.0_271]
at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) [?:1.8.0_271]
at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) [?:1.8.0_271]
at java.lang.Thread.run(Unknown Source) [?:1.8.0_271]

TheCiROMG:
Same problem T.T

Kamilkime:
I've made a temporary fix for that, working only on 1.16.4 - you can find it here:
https://github.com/games647/FastLogin/issues/412#issuecomment-721921736

games647:
Duplicate of #412

Issue #414, opened 2020-11-05
Couldn't enable encryption


kristupas098:
[//]: # (Lines in this format are considered as comments and will not be displayed.)
[//]: # (Before reporting make sure you're running the **latest build** of the plugin and checked for existing issues!)

NOTE: in spigot version you can join to server but in cracked mode, in paper nobody can join

### What behaviour is observed:
[//]: # (What happened?)
Premium players cannot login to server (disconnected)
Cracked player can login to server

### What behaviour is expected:
[//]: # (What did you expect?)
To join to server with premium account

### Steps/models to reproduce:
[//]: # (The actions that cause the issue. Please explain it in detail)
Update fastlogin to version 1.11-SNAPSHOT-f2441ec and spigot to Spigot-37d799b-3eb7236 or paper-git-261

### Screenshots (if applicable)
[//]: # (You can drop the files here directly)

### Plugin list:
[//]: # (This can be found by running `/pl`)
Authme, ChatCo, CustomHelpMessage, FastLogin, InfoBook, JoinMessage, KISS-tpa, LuckPerms, NoCheatPlus, OldschoolKill, ProtocolLib, RandomSpawn, RealPing, Rules, ServerListPlus, SkinsRestorer

### Environment description
[//]: # (Server software with exact version number, Minecraft version, SQLite/MySQL/MariaDB, ...)
git-Spigot-37d799b-3eb7236, Minecraft 1.16.4, paper-git-261

### Plugin version or build number (don't write latest):
[//]: # (This can be found by running `/version plugin-name`.)
1.11-SNAPSHOT-f2441ec

### Server Log:
[//]: # (No images please - only the textual representation)
[Hastebin](https://hastebin.com/) / [Gist](https://gist.github.com/) link of the error, stacktrace or the complete log (if any)
https://hastebin.com/bejadatamu.yaml

### Configuration:
[//]: # (No images please - only the textual representation)
[//]: # (remember to delete any sensitive data)
[Hastebin](https://hastebin.com/) / [Gist](https://gist.github.com/) link of your config.yml file 
https://hastebin.com/ocupasoniy.yaml

kristupas098:
NOTE: in papermc-256 puts error related to networkmanager

games647:
Duplicate #412

PR #418, opened 2020-11-08
Update encryption to handle MC 1.16.4, fixes #412

### LINKED ISSUE #412: Cannot send encryption packet. Falling back to cracked login
[//]: # (Before reporting make sure you're running the **latest build** of the plugin and checked for existing issues!)

### What behaviour is observed:
[//]: # (What happened?)
In version 1.16.4 then you login to server it asks your password.
Plugin uses cracked player info instead of premium user

### What behaviour is expected:
[//]: # (What did you expect?)
To login without password and using premium account

### Steps/models to reproduce:
[//]: # (The actions that cause the issue. Please explain it in detail)
Only need to update server software (spigot,papermc) to minecraft version 1.16.4

### Screenshots (if applicable)
[//]: # (You can drop the files here directly)

### Plugin list:
[//]: # (This can be found by running `/pl`)
Authme, chatco, customhelpmessage, fastchunkpregenerator, fastlogin, infobook ,joinmessage, kiss-tpa, luckperms, nocheatplus ,oldschoolkill ,protocollib, randomspawn, realping, rules, serverlistplus, skinsrestorer

### Environment description
[//]: # (Server software with exact version number, Minecraft version, SQLite/MySQL/MariaDB, ...)
Minecraft 1.16.4, papermc git-256

### Plugin version or build number (don't write latest):
[//]: # (This can be found by running `/version plugin-name`.)
1.11-snapshot-d034e71

### Server Log:
[//]: # (No images please - only the textual representation)
[Hastebin](https://hastebin.com/) / [Gist](https://gist.github.com/) link of the error, stacktrace or the complete log (if any)
https://hastebin.com/keqemahodi.yaml

### Configuration:
[//]: # (No images please - only the textual representation)
[//]: # (remember to delete any sensitive data)
[Hastebin](https://hastebin.com/) / [Gist](https://gist.github.com/) link of your config.yml file 
https://hastebin.com/ozacitojuz.yaml

Kamilkime:
### Summary of your change
On MC 1.16.3 and older - the encryption method would still be used like before
If the method is not found - it means it's MC 1.16.4 or older, the new encryption method will be used

### Related issue
Fixes #412 (and duplicates - #414, #416, #417)

### DIFF
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
@@ -175,16 +175,20 @@ public class VerifyResponseTask implements Runnable {
     }
 
     private boolean enableEncryption(SecretKey loginKey) throws IllegalArgumentException {
+        // Initialize method reflections
         if (encryptMethod == null) {
             Class<?> networkManagerClass = MinecraftReflection.getNetworkManagerClass();
 
             try {
+                // Try to get the old (pre MC 1.16.4) encryption method
                 encryptMethod = FuzzyReflection.fromClass(networkManagerClass)
                         .getMethodByParameters("a", SecretKey.class);
             } catch (IllegalArgumentException exception) {
+                // Get the new encryption method
                 encryptMethod = FuzzyReflection.fromClass(networkManagerClass)
                         .getMethodByParameters("a", Cipher.class, Cipher.class);
 
+                // Get the needed Cipher helper method (used to generate ciphers from login key)
                 Class<?> encryptionClass = MinecraftReflection.getMinecraftClass("MinecraftEncryption");
                 cipherMethod = FuzzyReflection.fromClass(encryptionClass)
                         .getMethodByParameters("a", int.class, Key.class);
@@ -194,13 +198,17 @@ public class VerifyResponseTask implements Runnable {
         try {
             Object networkManager = this.getNetworkManager();
 
+            // If cipherMethod is null - use old encryption (pre MC 1.16.4), otherwise use the new cipher one
             if (cipherMethod == null) {
+                // Encrypt/decrypt packet flow, this behaviour is expected by the client
                 encryptMethod.invoke(networkManager, loginKey);
             } else {
-                Object cipher = cipherMethod.invoke(null, 2, loginKey);
-                Object cipher1 = cipherMethod.invoke(null, 1, loginKey);
+                // Create ciphers from login key
+                Object decryptionCipher = cipherMethod.invoke(null, Cipher.DECRYPT_MODE, loginKey);
+                Object encryptionCipher = cipherMethod.invoke(null, Cipher.ENCRYPT_MODE, loginKey);
 
-                encryptMethod.invoke(networkManager, cipher, cipher1);
+                // Encrypt/decrypt packet flow, this behaviour is expected by the client
+                encryptMethod.invoke(networkManager, decryptionCipher, encryptionCipher);
             }
         } catch (Exception ex) {
             disconnect("error-kick", false, "Couldn't enable encryption", ex);
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
@@ -13,22 +13,23 @@ import com.github.games647.craftapi.model.skin.SkinProperty;
 import com.github.games647.craftapi.resolver.MojangResolver;
 import com.github.games647.fastlogin.bukkit.BukkitLoginSession;
 import com.github.games647.fastlogin.bukkit.FastLoginBukkit;
+import org.bukkit.Bukkit;
+import org.bukkit.entity.Player;
 
+import javax.crypto.Cipher;
+import javax.crypto.SecretKey;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.security.GeneralSecurityException;
+import java.security.Key;
 import java.security.KeyPair;
 import java.security.PrivateKey;
 import java.util.Arrays;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
```diff
@@ -175,13 +176,31 @@ private boolean enableEncryption(SecretKey loginKey) throws IllegalArgumentExcep
...
-            encryptMethod.invoke(networkManager, loginKey);
+            try {
+                //try to detect the method by parameters
+                Method encryptMethod = FuzzyReflection
+                        .fromObject(networkManager).getMethodByParameters("a", SecretKey.class);
+
+                //encrypt/decrypt following packets
+                //the client expects this behaviour
+                encryptMethod.invoke(networkManager, loginKey);
+            } catch (Exception ex) {
```
games647, line 187: Falling back on exceptions implies a very bad performance in my experience. The login process could be a very performance critical part, so it's the best to check the method availability once and then save the result. The best timing would be the initialization of the ProtocolLib module in my opinion.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
```diff
@@ -175,13 +176,31 @@ private boolean enableEncryption(SecretKey loginKey) throws IllegalArgumentExcep
...
+                //try to detect the method by parameters
+                Method encryptMethod = FuzzyReflection
+                        .fromObject(networkManager).getMethodByParameters("a", Cipher.class, Cipher.class);
+
+                //change Key instance into two Cipher instances
+                String ver = Bukkit.getServer().getClass().getPackage().getName().split("\\.")[3];
+                Class<?> encryptionClass = Class.forName("net.minecraft.server." + ver + ".MinecraftEncryption");
```
games647, line 194: ProtocolLib has a helper for finding the NMS class. Then we could rely on the unit tests provided by ProtocolLib.

games647 (changes requested):
Thanks for the help. However I have some minor suggestions.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
Kamilkime: I've switched to using MinecraftReflection class

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
Kamilkime: I've changed it so that the method reflections are static, so they will be initialized just once and reused

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
```diff
@@ -171,17 +175,33 @@ private Object getNetworkManager() throws IllegalAccessException, ClassNotFoundE
...
-            Object networkManager = getNetworkManager();
+            Object networkManager = this.getNetworkManager();
 
-            //try to detect the method by parameters
-            Method encryptMethod = FuzzyReflection
-                    .fromObject(networkManager).getMethodByParameters("a", SecretKey.class);
+            if (cipherMethod == null) {
+                encryptMethod.invoke(networkManager, loginKey);
+            } else {
+                Object cipher = cipherMethod.invoke(null, 2, loginKey);
```
games647, line 200: Please give them a more accurate name like how they are used.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
```diff
@@ -171,17 +175,33 @@ private Object getNetworkManager() throws IllegalAccessException, ClassNotFoundE
...
+
+            try {
+                encryptMethod = FuzzyReflection.fromClass(networkManagerClass)
+                        .getMethodByParameters("a", SecretKey.class);
+            } catch (IllegalArgumentException exception) {
+                encryptMethod = FuzzyReflection.fromClass(networkManagerClass)
+                        .getMethodByParameters("a", Cipher.class, Cipher.class);
+
+                Class<?> encryptionClass = MinecraftReflection.getMinecraftClass("MinecraftEncryption");
+                cipherMethod = FuzzyReflection.fromClass(encryptionClass)
```
games647, line 189: Please put in a comment what those two methods actually do. So the first one enables encryption, while the other initializes the cipher.

games647 (changes requested):

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
Kamilkime: I usually don't comment the code much - comments added now

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
Kamilkime: I assume you've meant the names and/or modes of Ciphers - I've changed both to make it all clearer

games647 (approved):

games647:
Perfect now, thanks

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
games647: In this context it was really clear what it does, so it was really necessary.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
games647: I meant the variable name (something like `initializedCipher`) and/or a description of what the method does, but it's even better now.

PR #443, opened 2021-01-12
Add UUID Mojang Player

DoctaEnkoda:
With the mojang API, determine the UUID of premium players


games647:
What is your motivation, what do you want to achieve? FastLogin already has an API for retrieving the UUID. Furthermore your current implementation would block the main thread, because you are performing a web requesting during command processing.

DoctaEnkoda:
Currently in the database, it remains a void, I was wrong of the place when requesting, there it should be good.

games647:
>Currently in the database, it remains a void,

That's normal, it will only be saved for premium players in order to reduce the number of UUID requests.

> I was wrong of the place when requesting, there it should be good.

Blocking requests should always run async. After completion it could the switch context and run again on the main thread.

PR #454, opened 2021-02-03
Implement BungeeCordAuthenticator

xXSchrandXx:

### Summary of your change
Modified pom.xml - Add SpigotPlugins-Repo
Modified pom.xml - Add SpigotPlugins do <excludes/>

Modified onEnable() - Register BungeeCordAuthenticator
Added BungeeCordAuthenticatorHook

### Related issue


xXSchrandXx:
Please excuse the many failed attempts due to the checksum from my spigotplugins-repo, I thought that this check could be ignored from maven...
I gave up and set it like the BungeeAuth dependency.

games647:
No need for any excuses. It failed, because checksum didn't match. You can see that the uploaded pom.sha1 indicates a different one to the pom file. I guess this has to that the file you uploaded has Linux file endings (git convention), but the project you built has different line endings (i.e. Windows). `git.autocrlf` modifies this behavior.

Normally it's bad practise to put binary objects in Git, because Git always preserves the history even if you deleted the file. BungeeAuth, is provided that way, because there was no other good way back then and it's unlikely we are going to replace it often. Nowadays GitHub provides an better way. You can use GitHub as Maven repository directly, without using the work-around source repository. 

https://docs.github.com/en/packages/guides/configuring-apache-maven-for-use-with-github-packages

Could you maybe try this out ?

xXSchrandXx:
I'll try that in the next days. 👍

bungee/pom.xml
```diff
@@ -30,6 +30,7 @@
                             <!--Those classes are already present in BungeeCord version-->
                             <exclude>net.md-5:bungeecord-config</exclude>
                             <exclude>com.google.code.gson:gson</exclude>
+                            <execlude>de.xxschrandxx.bca:BungeeCordAuthenticator</execlude>
```
games647, line 33: BungeeAuth is a system dependency. System and Provided repositories don't need the exclude tag.

games647 (changes requested):

xXSchrandXx:
Do you know how I can give everyone read access on the [package](https://github.com/xXSchrandXx/SpigotPlugins/packages/611440)?

xXSchrandXx:
Or do I just have to use this, too:
``` XML
        <repository>
          <id>central</id>
          <url>https://repo1.maven.org/maven2</url>
          <releases><enabled>true</enabled></releases>
          <snapshots><enabled>true</enabled></snapshots>
        </repository>
```

games647:
Okay wow, this makes it totally useless. They never mention this limitation. I thought public repositories also have public packages. I could provide an access token for GitHub actions, but it would still complicate the setup for others. https://jitpack.io/ would still be an option.

xXSchrandXx:
> Okay wow, this makes it totally useless. They never mention this limitation. I thought public repositories also have public packages. I could provide an access token for GitHub actions, but it would still complicate the setup for others. https://jitpack.io/ would still be an option.

I think they don't realy have this ready for every user. I don't even know how to delete this packages now 😄 ...

xXSchrandXx:
Woher kommt denn jetzt simonsators-repo (https://simonsator.de/repo) ??

xXSchrandXx:
I don't know why it doesn't use my repo...

I normally have it right... (browse for yourself: https://maven.gamestrike.de/mvn/)

games647:
The Let's Encrypt certificate could be too new for the Java 8 installation on Ubuntu 18.04. Normally GitHub announced that they are rolling out to 20.04, but we don't have it yet. You could change it manually: https://github.com/games647/FastLogin/blob/9ff404a424b20154108774050057651fb0ed4b03/.github/workflows/maven.yml#L15

If that doesn't work, it could be the old Java installation too, so a change to Java 11 is also acceptable. BTW JitPack normally expect one source repository per project, which is the normal convention. So you could have something like this: https://jitpack.io/#Steveice10/MCProtocolLib

Nevertheless your own Maven repository is also fine.

xXSchrandXx:
> The Let's Encrypt certificate could be too new for the Java 8 installation on Ubuntu 18.04. Normally GitHub announced that they are rolling out to 20.04, but we don't have it yet. You could change it manually:
> 
> https://github.com/games647/FastLogin/blob/9ff404a424b20154108774050057651fb0ed4b03/.github/workflows/maven.yml#L15
> 
> If that doesn't work, it could be the old Java installation too, so a change to Java 11 is also acceptable. BTW JitPack normally expect one source repository per project, which is the normal convention. So you could have something like this: https://jitpack.io/#Steveice10/MCProtocolLib
> 
> Nevertheless your own Maven repository is also fine.

But even http:// doesn't work?

xXSchrandXx:
I have the wrong name...

games647:
> I don't know why it doesn't use my repo...
> 
> I normally have it right... (browse for yourself: https://maven.gamestrike.de/mvn/)

It comes from your parent pom. Maven dependencies are transitive. So importing your plugin, imports all your dependencies as well.

games647:
You can set an exclude policy when specifying the dependency.

xXSchrandXx:
I had my artifactId in the dependency wrong because i had to lowecase it for github. And I forgot to change that back.
Now everything works

.gitignore
```diff
@@ -42,3 +42,6 @@ hs_err_pid*
 # Mac filesystem dust
 .DS_Store
 
+#lib
+#lib/
+.factorypath
```
games647, line 47:
```suggestion

```

Please remove it if it's unused and end file with an empty line

bungee/pom.xml
```diff
@@ -65,6 +66,12 @@
             <id>nukkitx-repo</id>
             <url>https://repo.nukkitx.com/maven-snapshots/</url>
         </repository>
+
+		<repository>
+		    <id>spigotplugins-repo</id>
+		    <url>https://maven.gamestrike.de/mvn/</url>
+		</repository>
+
```
games647, line 74: Wrong indentation

```suggestion

        <repository>
                <id>spigotplugins-repo</id>
                <url>https://maven.gamestrike.de/mvn/</url>
        </repository>
```

bungee/src/main/java/com/github/games647/fastlogin/bungee/FastLoginBungee.java
```diff
@@ -84,11 +85,17 @@ public void onDisable() {
     }
 
     private void registerHook() {
-        Plugin plugin = getProxy().getPluginManager().getPlugin("BungeeAuth");
-        if (plugin != null) {
+        Plugin plugin1 = getProxy().getPluginManager().getPlugin("BungeeAuth");
```
games647, line 88: Please give both plugin variables a more meaningful name. Otherwise mistakes could happen easily. Something like `bungeAuthPlugin` and `authenticatorPlugin`.

games647 (changes requested):
Only some minor things

games647:
You also need to remove the binary file. It's no longer used.

games647 (approved):

Issue #458, opened 2021-02-08
AutoRegister Bypass 


elmaslarry:
[//]: # (Lines in this format are considered as comments and will not be displayed.)
[//]: # (Before reporting make sure you're running the **latest build** of the plugin and checked for existing issues!)

### What behaviour is observed:
[//]: # I have the setting: autoregister: true enabled, and my players have found a bypass. Through authme I have the account limit per IP at 2, however with this autoregister fastlogin setting, players can register infinite accounts on the server taking advantage of several multi-accounts, I have come to see players who have up to 30 registered accounts. I would like this to be solved as soon as possible. Thank you!

### What behaviour is expected:
[//]: # That the fastlogin plugin does a check of the accounts registered in that IP before allowing the registration of a new account, to avoid bypassing the limit of registrations in authme

### Steps/models to reproduce:
[//]: # autoRegister: true  in config of fastlogin

### Screenshots (if applicable)
[//]: # Don't have

### Plugin list:
[//]: # it's not important

### Environment description
[//]: # it's not important

### Plugin version or build number (don't write latest):
[//]: # 959

### Server Log:
[//]: # It's not important

### Configuration:
[//]: # autoRegister: true

Issue #457, opened 2021-02-07
nameChangeCheck: true [Broken]


WalrusSlayer:
[//]: # (Lines in this format are considered as comments and will not be displayed.)
[//]: # (Before reporting make sure you're running the **latest build** of the plugin and checked for existing issues!)

### What behaviour is observed:
[//]: # (What happened?)
nameChangeCheck: true 
Does not work.
### What behaviour is expected:
[//]: # (What did you expect?)
My issue is that when a premium user, for instance: "Anthony555" changes his name to "DogLover115", it treats doglover115 as a different account even though its the same premium user. Based on the info in the config, it should automatically sign you in if you change your name and are a premium user. Everything else in the plugin works flawlessly.
### Steps/models to reproduce:
[//]: # (The actions that cause the issue. Please explain it in detail)
Simply enable nameChangeCheck: true in the config, and have a premium account change its name. It won't auto log them in. Instead, it treats them as a separate cracked user. even when premium UUID's are enabled as well.
### Screenshots (if applicable)
[//]: # (You can drop the files here directly)
![image](https://user-images.githubusercontent.com/40503028/107146553-1573b180-6917-11eb-98a5-45a19249765e.png)
### Plugin list:
[//]: # (This can be found by running `/pl`)
![image](https://user-images.githubusercontent.com/40503028/107146841-91bac480-6918-11eb-84b0-2753680651f7.png)
### Environment description
[//]: # (Server software with exact version number, Minecraft version, SQLite/MySQL/MariaDB, ...)
PaperMC #465 (1.16.5) Waterfall #395, MySQL.
### Plugin version or build number (don't write latest):
[//]: # (This can be found by running `/version plugin-name`.)
1.11-SNAPSHOP-9ff404a
### Server Log:
[//]: # (No images please - only the textual representation)
Paper: https://paste.gg/p/Walrus2020/ba5f3eba66204f6aa6304c2bb6d6e7f9
Waterfall: https://paste.gg/p/Walrus2020/d393622969a4435f8675cec3cac57f5c
### Configuration:
[//]: # (No images please - only the textual representation)
[//]: # (remember to delete any sensitive data)
https://paste.gg/p/Walrus2020/ad08b1a9fed447e59e750c025c742765

Kamilkime:
It's not a perfect solution, but I've managed to 'repair' it in #437

WalrusSlayer:
> It's not a perfect solution, but I've managed to 'repair' it in #437

[...]

Issue #437, opened 2021-01-09
FastLogin forcing old username (nameChangeCheck not working)


Kamilkime:
[//]: # (Lines in this format are considered as comments and will not be displayed.)
[//]: # (Before reporting make sure you're running the **latest build** of the plugin and checked for existing issues!)

### What behaviour is observed:
After a player changed their premium username - FastLogin still forces an old username to be used. Also because of that - premium register/login is not working for that person and he has to log in manually

### What behaviour is expected:
FastLogin should recognize the change, update the database and not force the old username

### Steps/models to reproduce:
Login onto the server before change, change the premium username, log in again

### Plugin list:
```
AnimatedScoreboard, AuthMe, ChestShop, Chunky, CoreProtect, CraftBook, CraftserveAgent*, CustomFishing*,
EntityCounter, Essentials, EssentialsChat, EssentialsGeoIP, FastAsyncWorldEdit (WorldEdit), FastLogin, LWC,
MenuBuilder*, PermissionsEx*, PlaceholderAPI, PlayerHeads, ProtocolLib, SelectionVisualizer, SkinsRestorer,
SpigotTest*, TAB, TempTitanBans, TempTitanChat, TempTitanCommands, TempTitanCore, TempTitanDistance,
TempTitanEconomy, TempTitanMySQL, TempTitanRegions, TempTitanSaddles, TempTitanSleep, TempTitanTickets,
TempTitanVote, TempTitanWhitelist, TitanTempMotd, TreeGravity, Vault, WorldGuard
```

### Environment description
Paper 1.16.4, latest FastLogin dev-build with MySQL

### Server Log:
```
[20:19:57] [FastLogin Pool Thread #1/INFO]: [FastLogin] Handling player xMandzio
[20:19:58] [FastLogin Pool Thread #1/INFO]: [FastLogin] GameProfile Profile{id=4e829203-7bb1-49ca-8f2a-5a03c099c3d1, name='xMandzio'} changed it's username
[20:19:58] [FastLogin Pool Thread #2/INFO]: [FastLogin] Profile xMandzio has a verified premium account: Verification{id=4e829203-7bb1-49ca-8f2a-5a03c099c3d1, name='xMandzio', properties=[SkinProperty{value='longbase64'}]}
[20:19:58] [User Authenticator #49/INFO]: UUID of player MinecraftPapryka is 4e829203-7bb1-49ca-8f2a-5a03c099c3d1
[...]

PR #459, opened 2021-02-09
Fix name change detection, allow FastLogin to respect AuthMe registration limits

### LINKED ISSUE #423: Premium username change -> few "problems".
**Hi!**

**I use FastLogin with AdvancedLogin.**
FastLogin config: https://pastebin.com/QU3bn52S
AdvancedLogin (not very useful...): https://pastebin.com/29PUh85j

It's quite new on my server, as I began to have more and more players, and wanted something better for the players that changed their username. My server runs on OnlineMode=false.

One of my players changed his username from Jobeee to mvjik. AdvancedLogin saw that, did the change, and handles him as a "cracked"... didn't really understood why. So now when we parse %fastlogin_status% on him it returns "Unknown".

So he has to enter his passcode, and second issue, he is still considerated as Jobeee....?..... I don't know what I did wrong, but I did. When he write on the chat it is Jobeee, when we look on the Dynmap it is jobeee, but if we want to change permission (LuckPerms) then it is mvjik!....

Hope you will be able to help me. Thanks

Kamilkime:
### Summary of your change
First change fixes name change detection on PaperSpigot. Because the gameprofile didn't have the skin set and the server is set to offline-mode - Paper was using a cached profile from usercache.json, which had the old username. That way the name change was spotted, but basically ignored by Paper.

Second change allows to respect the per IP registration limit of AuthMe. Due to the limitations of its API - some reflections have to be used. When registering using the API - a different type of registration is triggered. The API one doesn't check for IP limits. The reflection allows FastLogin to trigger a registration like it would be done using the `/register` command - that way the limits will be checked.

Also because AuthMe makes the registration async - there's no good way to detect if the registration was successful, FastLogin will always display a message about auto-register and then AuthMe will send a message about IP limit. I've changed the default auto-register message to at least a bit account for possible registration fail.

### Related issue
Fixes #423, #437, #457, #458

### DIFF
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/AuthMeHook.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/AuthMeHook.java
@@ -3,14 +3,17 @@ package com.github.games647.fastlogin.bukkit.hook;
 import com.github.games647.fastlogin.bukkit.BukkitLoginSession;
 import com.github.games647.fastlogin.bukkit.FastLoginBukkit;
 import com.github.games647.fastlogin.core.hooks.AuthPlugin;
-
 import fr.xephi.authme.api.v3.AuthMeApi;
 import fr.xephi.authme.events.RestoreSessionEvent;
-
+import fr.xephi.authme.process.Management;
+import fr.xephi.authme.process.register.executors.ApiPasswordRegisterParams;
+import fr.xephi.authme.process.register.executors.RegistrationMethod;
 import org.bukkit.entity.Player;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
 import org.bukkit.event.Listener;
+
+import java.lang.reflect.Field;
 
 /**
  * GitHub: https://github.com/Xephi/AuthMeReloaded/
@@ -25,8 +28,22 @@ public class AuthMeHook implements AuthPlugin<Player>, Listener {
 
     private final FastLoginBukkit plugin;
 
+    private final AuthMeApi authmeAPI;
+    private Management authmeManagement;
+
     public AuthMeHook(FastLoginBukkit plugin) {
         this.plugin = plugin;
+        this.authmeAPI = AuthMeApi.getInstance();
+
+        if (plugin.getConfig().getBoolean("respectIpLimit", false)) {
+            try {
+                Field managementField = this.authmeAPI.getClass().getDeclaredField("management");
+                managementField.setAccessible(true);
+                this.authmeManagement = (Management) managementField.get(this.authmeAPI);
+            } catch (NoSuchFieldException | IllegalAccessException exception) {
+                this.authmeManagement = null;
+            }
+        }
     }
 
     @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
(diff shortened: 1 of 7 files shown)


games647:
>Also because AuthMe makes the registration async - there's no good way to detect if the registration was successful, FastLogin will always display a message about auto-register and then AuthMe will send a message about IP limit. I've changed the default auto-register message to at least a bit account for possible registration fail.

Yeah it's a little bit sad, almost all auth plugins do not expose an thread-safe API which is also returns the success status. There are many work-arounds.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/AuthMeHook.java
```diff
@@ -41,25 +58,32 @@ public void onSessionRestore(RestoreSessionEvent restoreSessionEvent) {
...
-        return true;
+        //if we have the management - we can trigger register with IP limit checks
+        if (authmeManagement != null) {
+            authmeManagement.performRegister(RegistrationMethod.PASSWORD_REGISTRATION,
+                    ApiPasswordRegisterParams.of(player, password, true));
+        } else {
+            authmeAPI.forceRegister(player, password);
+        }
+
+        return authmeAPI.isRegistered(player.getName());
```
games647, line 87: Couldn't this cause an race-condition? If the calls are async, they would immediately return and therefore the check `authmeAPI.isRegistered` could return false even though the registration will be successful. 

This has to be tested with async enabled AuthMe.

core/src/main/resources/config.yml
```diff
@@ -36,6 +36,15 @@ anti-bot:
 # For more information: https://github.com/games647/FastLogin#why-do-players-have-to-invoke-a-command
 autoRegister: false
 
+# Should FastLogin respect per IP limit of registrations (e.g. in AuthMe)
+# Because most auth plugins do their stuff async - FastLogin will still think the player was registered
+# To work best - you also need to enable auto-register-unknown
```
games647, line 41: `auto-register-unknown` You meant `auto-register`?

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/AuthMeHook.java
Kamilkime: Yeah, I forgot to change that back to true - checking if isRegistered() sometimes works and sometime doesn't

core/src/main/resources/config.yml
Kamilkime: Nope, I meant `auto-register-unknown` - that way FastLogin will try to register the player again, so the player would get the message about the IP limit again. If `auto-register-unknown` is set to false - they would only get that message the first time they join and they might not see it immediately

If `auto-register` is set to false - the IP limit would just not be used by FastLogin, because it wouldn't register the player at all

games647 (approved):

games647:
Thank you very much.

Fargus5:
Thank you for working on it.
Hope that a solution will be update soon.
Thanks to both for what you are doing for all of us

elmaslarry:
Me gustaría saber si se arreglo el bypass con los registros de ip en authme con la ultima version? No quiero dejar que los jugadores puedan bypassear ese limite de registros por IP 

Si finalmente se arreglo, debo cambiar algo en mi configuracion? o simplemente reemplazar el -jar
Gracias

elmaslarry:
I would like to know if the bypass with the ip registers in authme with the latest version was fixed? I don't want to let players bypass that IP record limit

If it is finally fixed, should I change something in my configuration? or just replace the -jar
Thank you

Kamilkime:
> I would like to know if the bypass with the ip registers in authme with the latest version was fixed? I don't want to let players bypass that IP record limit
> 
> If it is finally fixed, should I change something in my configuration? or just replace the -jar
> Thank you
> 

Yeah, you need to add a **respectIpLimit** section to your config.yml and set it's value to true. I'd also advice enabling **auto-register-unknown** for the best experience

Fargus5:
@Kamilkime I saw that there was an exchange between you and the developer. the problem was solved or are you still working on it? I just ask you to know why no update has been released on spigot yet.
Sorry to bother you.

Kamilkime:
> @Kamilkime I saw that there was an exchange between you and the developer. the problem was solved or are you still working on it? I just ask you to know why no update has been released on spigot yet.
> Sorry to bother you.

The issue was fixed - but there just aren't any official releases on spigotmc. To get the newest version with the fixes - you need to get it from https://ci.codemc.org/job/Games647/job/FastLogin/changes

Fargus5:
thanks

elmaslarry:
So i need to put: 

respectIpLimit: true

autoRegister: true 


¿?

Kamilkime:
> So i need to put: 
> 
> respectIpLimit: true
> 
> autoRegister: true 
> 
> 
> ¿?
> 
> 

Yup, and it would be best to also enable auto-register-unknown

elmaslarry:
Okey thanks!

elmaslarry:
I need to add some new message ¿? in lang file ?

Kamilkime:
> I need to add some new message ¿? in lang file ?

Nope, the messages are all the same as before

Issue #328, opened 2020-04-22
Floodgate (plugin) support


Patbox:
[//]: # (Lines in this format are considered as comments and will not be displayed.)
Floodgate is a plugin, that adds better integration with Geyser (proxy).
### Is your feature request related to a problem? Please describe.
[//]: # (A clear and concise description of what the problem is. Ex. I'm always frustrated when [...])
When you setup flodgate, you can set prefix to bedrock player's names, so they will not overlap with java ones. But FastLogin prevents this functionality (aka no prefix)
### Describe the solution you'd like
[//]: # (A clear and concise description of what you want to happen.)
Add support to Floodgate/Make that FastLogin gets username after prefix is added.
### Describe alternatives you've considered
[//]: # (A clear and concise description of any alternative solutions or features you've considered.)
If you want use prefixes, there isn't any.
### Additional context
[//]: # (Add any other context or screenshots about the feature request here.)
https://github.com/GeyserMC/Floodgate
https://github.com/GeyserMC/Geyser

games647:
Do you specifically have this problem with BungeeCord or Spigot? BungeeCord shouldn't cause much issues, because it uses normal online mode connections. Only the offline UUID is set manually for online mode connections.

Patbox:
With Spigot. Also there is log, when player's login

`[12:41:58 INFO]: [Geyser-Bukkit] /127.0.0.1:45172 tried to connect!
[12:41:58 INFO]: [Geyser-Bukkit] Player connected with username TriadicPatbox
[12:42:01 INFO]: [Geyser-Bukkit] Attempting to login using floodgate mode... authentication will be encrypted
[12:42:01 INFO]: [Geyser-Bukkit] Loaded Floodgate key!
[12:42:01 INFO]: [Geyser-Bukkit] TriadicPatbox (logged in as: TriadicPatbox) has connected to remote java server on address 127.0.0.1
[12:42:01 INFO]: [floodgate-bukkit] Added $TriadicPatbox 00000000-0000-0000-0009-01fe4fcae617
[12:42:01 INFO]: UUID of player TriadicPatbox is 00000000-0000-0000-0009-01fe4fcae617
[12:42:01 INFO]: TriadicPatbox[/127.0.0.1:40598] logged in with entity id 2081 at ([world]239.7, 68.0, 390.2451)
[12:42:01 INFO]: » [Gracz] TriadicPatbox dołączył(a) do gry!
[...]

Issue #493, opened 2021-03-20
Floodgate player name prefixes are not working with ProtocolLib


Smart123s:
[//]: # (Lines in this format are considered as comments and will not be displayed.)
[//]: # (Before reporting make sure you're running the **latest build** of the plugin and checked for existing issues!)

### What behaviour is observed:
[//]: # (What happened?)
When I join the game, my name is Smart123s.
In the log: `Smart123s left the game`

If I delete `FastLogin.jar` from the `plugins` folder, the prefixes are added to the player's name as excepted.

### What behaviour is expected:
[//]: # (What did you expect?)
When I join the game, my name should be *Smart123s.
In the log: `*Smart123s left the game`

### Steps/models to reproduce:
[//]: # (The actions that cause the issue. Please explain it in detail)
Install the plugins mentioned below, and login from Bedrock Edition.

### Plugin list:
[//]: # (This can be found by running `/pl`)
AuthMe, FastLogin, Geyser-Spigot, floodgate-bukkit, ProtocolLib, Skinsrestorer

### Environment description
[//]: # (Server software with exact version number, Minecraft version, SQLite/MySQL/MariaDB, ...)
Default DB for all plugins.
Java 14, Ubuntu 20.04, MC v1.16.5, Paper-553, No Bungeecord or Proxy
Connected to server using an Android device.
The issue is present both on LAN and on Cellular

### Plugin version or build number (don't write latest):
[//]: # (This can be found by running `/version plugin-name`.)
[Build #968](https://ci.codemc.io/job/Games647/job/FastLogin/968/)

### Server Log:
[//]: # (No images please - only the textual representation)
None

### Configuration:
[//]: # (No images please - only the textual representation)
[//]: # (remember to delete any sensitive data)
*Note: other entries have not been modified*
```
autoRegister: false
secondAttemptCracked: true
switchMode: false
premiumUuid: true
nameChangeCheck: false
forwardSkin: true
premium-warning: true
mojang-request-limit: 600
auto-register-unknown: false
autoLogin: true
```

Floodgate's config:
```
username-prefix: "*"
```

Geyser's config:
```
remote:
    auth-type: floodgate
```

t0ssox:
This bug causes that a significant number of players (for my small server) cannot enter via bedrock because most have the same nickname on Bedrock and Java. Hope this will be fixed in the future.

Smart123s:
I've done some debugging and now I am even more confused than I was before.

[...]

PR #494, opened 2021-03-20
Floodgate support for Bukkit

Smart123s:

## Summary of your change
Adds support for automatically registering and logging in with both linked and standalone Bedrock Players. Auto login is verified by Floodgate and the player's Xbox Live account.

## New Configuration options
*These options only work on Bukkit/Spigot/Paper servers, not on BungeeCord*

### autoLoginFloodgate
This enables auto login for every player connecting through Floodgate.
Possible values: `false`, `true`, `linked`
Linked means that only Bedrock accounts linked to a Java account will be logged in automatically
*This value is treated as a string ~~and requires apostrophes~~*

### allowFloodgateNameConflict
This enables Floodgate players to join the server, even if autoRegister is true and there's an existing Java Premium
account with the same name
Possible values:
- **false**: Check for Premium Java name conflicts as described in [`autoRegister`](https://github.com/games647/FastLogin/blob/main/core/src/main/resources/config.yml#L32)
  -  `autoRegister` must be `true` for this to work
- **true**:  Bypass `autoRegister`'s name conflict checking 
- **linked**: Bedrock accounts linked to a Java account will be allowed to join with conflicting names
~~Note: Linking a new account requires players to log in with a non-linked Bedrock account first
Enabling this will make linking new Bedrock players impossible
More information on linking accounts: https://github.com/GeyserMC/Geyser/wiki/Floodgate#account-linking
Releated Floodgate issue: https://github.com/GeyserMC/Floodgate/issues/37~~
*Floodgate 2.0 uses a different Linking method (called Global Linking) and this is no longer relevant.*
*This value is treated as a string ~~and requires apostrophes~~*

### autoRegisterFloodgate
This enables auto registering every player connecting through Floodgate.
`autoLoginFloodgate` must be `true` for this to work

## Related issue
Implements / fixes https://github.com/games647/FastLogin/issues/328
Additional issues in the implementation may be caused by https://github.com/games647/FastLogin/issues/493

~~**There is an issue UNRELATED to this fork which makes the building of the Bungee part of the plugin fail. The fork can still be tested as it does NOT touch the Bungee version.** For more information, see [this build report](https://github.com/games647/FastLogin/runs/2132145498#step:5:1404) from the upstream version of the plugin.~~ This is now fixed.

## Known Bugs
- [x] ~~Auto Registering from a Bedrock clinet will cause a Java Client with the same name to also be registered, but will not be automatically logged in~~ Auto (and not auto?) registering with a Java account and then logging out, and logging in with a bedrock account with the same name will cause the Bedrock account to be auto logged in without linking. See also at [Tests - Same bedrock and java usernames](#same-bedrock-and-java-usernames).
- [x] Profile.isSaved() is 'false' when logging in from Bedrock after auto registering through Bedrock so FastLogin will try to register for a second time, insted of logging in -- fixed by switching to `authPlugin.isRegistered()`
- [x] (Auto?) Registering as a Bedrock player will not create/store a StoredProfile in the database.
- [x] Registering from a Bedrock account gives "Successfully registered", and right after that "This user isn't registered". Joining for a second time successfully logs the player in. So the problem is with logging in right after registering. Maybe it something to do with the asynchronous task? - EDIT (2020 March 25): It's no longer doing it. Nothing has been changed since the last test.
  
## To be tested
All of the points below should be tried with _almost_ every possible combinations of the configuration entries: `autoRegister`, `premiumUuid`, `nameChangeCheck`, `autoLogin`, `autoLoginFloodgate`, `allowFloodgateNameConflict`, `autoRegisterFloodgate`

### Same bedrock and java usernames
- [x] Does the Floodgate config `username-prefix` make FastLogin treat same named users as if they had different names? (https://github.com/games647/FastLogin/issues/493)
- [x] Does allowFloodgateNameConflict work?
#### With unlinked accounts - Same bedrock and java usernames
- [x] With the Java account logging in first (and registering)
  - [x] Auto login from Java account
  - [x] Auto login from Bedrock account
- [x] With the Bedrock account logging in first (and registering) _see known bugs_
  - [x] Auto login from Java account
  - [x] Auto login from Bedrock account
#### With linked accounts (after linking) - Same bedrock and java usernames
  - [x] Auto login from Java account
  - [x] Auto login from Bedrock account

### Different bedrock and java usernames
  - [x] Auto login from Bedrock account
  - [x] Auto register from Bedrock account
  - [x] Linked account auto login from both devices

### Miscellaneous
- [x] ~~Does the FloodgateAPI return null ([used here](https://github.com/Smart123s/FastLogin/blob/main/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/ConnectionListener.java#L63)) if Geyser config `remote.auth-type` is set to `offline` and Floodgate is installed on the server and...~~
- ~~...the player is using cracked Minecraft Bedrock (the official client does not allow you to play multiplayer without logging in to Xbox Live) - **IF NOT, THIS CAN ALLOW CRACKED MC-BE USERS TO AUTO-LOGIN WITH ANY USERNAME**~~
- ~~...the player is logged in to Xbox Live when connecting - _This should be fine_~~
**Added `GeyserConnector.getInstance().getDefaultAuthType() == AuthType.FLOODGATE` which should solve this problem.**

## To Do
*This Todo list is incomplete, please help me add elements to it.*
- [x] Find a better way to implement 'autoRegisterFloodgate: false' in [ConectionListener.java#98](https://github.com/Smart123s/FastLogin/blob/e1af75191edf9bce5e9d9b441b85b36f535eee0f/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/ConnectionListener.java#L98)
- [x] Rewrite comments in [config.yml](https://github.com/Smart123s/FastLogin/blob/main/core/src/main/resources/config.yml) to make them easier to understand
  - [x] Add experimental warning messages
- [x] Check if the workaround for used to detect players connecting through Geyser is reliable ([use1](https://github.com/Smart123s/FastLogin/blob/e1af75191edf9bce5e9d9b441b85b36f535eee0f/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/ConnectionListener.java#L58), [use2](https://github.com/Smart123s/FastLogin/blob/a23f846146358860accdd7514285306e25692517/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/NameCheckTask.java#L49))
EDIT: the use cases point to the removed Floodgate 1.0 implementation.
  - [x] Maybe put them into a common function to avoid code duplication?
- [x] Decide if players with the same name should be auto logged in from different platforms (they have different UUIDs)
This behavior will be made configurable in a future Pull Request.
- [x] Support [Floodgate 2.0](https://github.com/GeyserMC/Floodgate/tree/dev/2.0/)
- [ ] Check if the changes didn't break Floodgate's Xbox authentication. Floodgate should block offline Bedrock players from connecting.


Smart123s:
There is an issue **unrelated** to this fork which makes the building of the Bungee part of the plugin fail. The fork can still be tested as it does not touch the Bungee version. For more information, see [this build report](https://github.com/games647/FastLogin/runs/2132145498#step:5:1404) from the upstream version of the plugin.

games647 (changes requested):
> For more information, see this build report from the upstream version of the plugin.

I noticed that. I hoped the author would fix it. I'll go to look into an alternative. 

>Auto Registering from a Bedrock clinet will cause a Java Client with the same name to also be registered, but will not be automatically logged in

The ForceLoginTask will register and login the player at the same time. So I wonder what went wrong here or are they separate login events?

> Profile.isSaved() is 'false' when logging in from Bedrock after auto registering through Bedrock so FastLogin will try to register for a second time, insted of logging in

Try the auth plugin isRegstered method

> Maybe put them into a common function to avoid code duplication?

You could also put the stuff into an extra class in order to isolate it. This could make it easier to test against it.

bukkit/pom.xml
```diff
@@ -213,6 +219,20 @@
                 </exclusion>
             </exclusions>
         </dependency>
+
```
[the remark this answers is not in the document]
Smart123s: I've moved the dependencies above `<!--Login Plugins-->` and left the repositories alone.
Please check if they in the right place now.

Smart123s:
> > For more information, see this build report from the upstream version of the plugin.
> 
> I noticed that. I hoped the author would fix it. I'll go to look into an alternative.

I have opened [another Pull Request](https://github.com/games647/FastLogin/pull/496) that could be an alternative to that.

Smart123s:
>> Maybe put them into a common function to avoid code duplication?
>
> You could also put the stuff into an extra class in order to isolate it. This could make it easier to test against it.

Can I create a `FloodgateHook` class in `com.github.games647.fastlogin.bukkit.hook` or is it reserved for auth plugins?
Or should I put it somewhere else?

games647:
>Can I create a FloodgateHook class in com.github.games647.fastlogin.bukkit.hook or is it reserved for auth plugins?
Or should I put it somewhere else?

That's normally for auth plugins. Maybe we should choose a better name for that. You could put into the auth package, because it's related to login/auth handling.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/ConnectionListener.java
```diff
@@ -45,6 +50,57 @@ public void onPlayerJoin(PlayerJoinEvent joinEvent) {
...
+            
+            if(Bukkit.getServer().getPluginManager().getPlugin("Geyser-Spigot") != null &&
+                    Bukkit.getServer().getPluginManager().getPlugin("floodgate-bukkit") != null) {                
+                FloodgatePlayer floodgatePlayer = null;
+                
+                // check if the player is really connected through Geyser
+                for (GeyserSession geyserPlayer : GeyserConnector.getInstance().getPlayers()) {
+                    if (geyserPlayer.getName().equals(player.getName())) {
+                        // this also returns a floodgatePlayer for linked Java accounts
+                        // that's why the Geyser Server's player list also has to be checked
```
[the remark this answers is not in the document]
Smart123s: According to my tests, `FloodgateAPI.getPlayer(player.getUniqueId())` also returns a `FloodgatePlayer` instance if the player is **connecting from Java Edition** (JE) but has a linked Bedrock Edition (BE) account.
Later in the code `floodgatePlayer != null` is used to decide if a player is connecting from BE. But since JE accounts with linked BE accounts also have a return value, I had to filter them out. That's what the for loop is used for.
I couldn't find a better solution (yet), but honestly, I didn't spend much time looking. I am planning to figure out an alternative method for this. If I won't find anything better, than at least I'll try switching `name` to `uuid` in [line 60](https://github.com/games647/FastLogin/pull/494/files/e1af75191edf9bce5e9d9b441b85b36f535eee0f#diff-554627ddde4f64e3c694783b936aef722dc908f261e03813733df48483e39f49R60).

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/ConnectionListener.java
```diff
@@ -45,6 +50,57 @@ public void onPlayerJoin(PlayerJoinEvent joinEvent) {
...
+                    }
+                }
+                
+                if (floodgatePlayer != null) {
+                    plugin.getLog().info(
+                            "Player {} is connecting through Geyser Floodgate.",
+                            player.getName());
+                    String allowNameConflict = plugin.getCore().getConfig().getString("allowFloodgateNameConflict");
+                    // check if the Bedrock player is linked to a Java account 
+                    boolean isLinked = floodgatePlayer.fetchLinkedPlayer() != null;
```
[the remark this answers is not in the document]
Smart123s: I've moved this to an asynchronous task.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/ConnectionListener.java
```diff
@@ -45,6 +50,57 @@ public void onPlayerJoin(PlayerJoinEvent joinEvent) {
...
+
+                    }
+                    if (!allowNameConflict.equalsIgnoreCase("true") && !allowNameConflict.equalsIgnoreCase("linked")) {
+                        plugin.getLog().error(
+                                "Invalid value detected for 'allowNameConflict' in FasttLogin/config.yml. Aborting login of Player {}",
+                                player.getName());
+                        return;
+                    }
+                    
+                    StoredProfile profile = plugin.getCore().getStorage().loadProfile(player.getName());
```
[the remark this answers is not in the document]
Smart123s: This has been removed in favor of `authPlugin.isRegistered()` which has also been moved to an asynchronous task.

Smart123s:
I need your opinion on this one _(and on everything else of course)_:
In this scenario there are two players with the same name, but one of them is playing from Java Edition (JE), and the other one is playing from Bedrock Edition (BE).
They obviously can't be online at the same time, because one name can only be online at a time on the server. (Who would /tp apply to if there were two Steves on the server? :smile: )
However, if they play at different times, than I've allowed auto login for both of them (if auto-login is enabled for both playtforms in the config, of course), since they have different UUIDs. I have tested this with the following combinations:
- Cracked JE - Bedrock (with Xbox)
- Premium JE - Bedrock (with Xbox)
Floodgate requires Xbox, so disabling it is not an option.

Is it good this way, or should only one of them be able to auto log in?
Note 1 : Floodgate prefixes are not working (https://github.com/games647/FastLogin/issues/493), which would eliminate this issue _for servers who are using them._
Note 2: The `allowFloodgateNameConflict` configuration option only cehcks for Premium Java Usernames at the moment, using [NameCheckTask](https://github.com/Smart123s/FastLogin/blob/main/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/NameCheckTask.java#L52).

Smart123s:
Okay, so I think that the PR is ready _(well, besides a new bug that I've added to the Known bugs list, any idea on how to fix it?)_ for another review, and for merging if you find everything alright. Could you please check the tasks that I've left unticked? Also, I've commented on all of your "change requests", could you close them if you are satisfied with my solutions? Please, don't forget to properly **test my code** before merging it, I'm not sure if I hadn't broken anything, and I don't want to compromise anyone's server's security. Also, are the configuration options understandable?

games647:
Testing requires some time. I'll look into it the next days.

bobhenl:
And please add ver 2.0 support too

Smart123s:
> And please add ver 2.0 support too

What do you mean by ver 2.0? I know that it was mentioned in https://github.com/games647/FastLogin/issues/328#issuecomment-679353587 as an upcoming release, but I couldn't find it anywhere. Could you please provide a link to ver 2.0?
I've worked with this one: https://github.com/GeyserMC/Floodgate/

bobhenl:
I've opened ticket https://github.com/games647/FastLogin/issues/498 and mentioned, that on ver 2.0 autologin doesn't work... on Floodgate 1.0 it normally works. https://ci.opencollab.dev/job/GeyserMC/job/Floodgate/job/dev%252F2.0/

Smart123s:
I am going to rewrite the PR to support Floodgate 2.0. With that, I'll drop support for version 1.x. The API of 2.0 just seem to be more polished at first glance.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
```diff
@@ -0,0 +1,93 @@
...
+    private final Player player;
+
+    public FloodgateAuthTask(FastLoginBukkit plugin, Player player) {
+        this.plugin = plugin;
+        this.player = player;
+    }
+
+    @Override
+    public void run() {
+        FloodgatePlayer floodgatePlayer = FloodgateApi.getInstance().getPlayer(player.getUniqueId());
```
games647, line 25: Please re-use the floodgate player, that saves us an additional lookup.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
```diff
@@ -0,0 +1,93 @@
...
+            // kicking must be synchronous
+            // https://www.spigotmc.org/threads/asynchronous-player-kick-problem.168580/
+            Bukkit.getScheduler().runTask(plugin, new Runnable() {
+                public void run() {
+                    player.kickPlayer("This name is allready in use by a Premium Java Player");
+                }
+            });
+
+        }
+        if (!allowNameConflict.equalsIgnoreCase("true")
```
games647, line 46: Put this into an extra method. It's a little too complicated.

games647 (changes requested):

games647:
@Smart123s So should I start to test it?

Smart123s:
I don't think that I could do anything other than to test it, so It'd be cool if you could test it too. To be honest, I didn't have much time to test with v2 so you might come across some issues.

Smart123s:
By the way @games647, the MIT license states that 

> The above copyright notice and this permission notice shall be included in all
copies or **substantial portions** of the Software.

So, could you please add the copyright headers to every _substantial portion_ (aka. file) of the plugin before merging this PR?
I really don't want to get sued because I didn't apply an appropriate no-warranties header in the source code of a Minecraft plugin.
For an example the copyright headers are present in [Floodgate's source](https://github.com/GeyserMC/Floodgate/blob/ab4696a3609248525ac9483eabb2a540272c6509/spigot/src/main/java/org/geysermc/floodgate/SpigotPlugin.java).
I've just realized this, I hope it's not a too big request.

games647:
@Smart123s That's only necessary if you include the code. We are only using the API and calling it. It is something else if you shade/copy the project. If you for example have a plugin (you could also take an Android app as an example) and shade a library into the end product, then you have to specify the license.

Smart123s:
My example in the end of my last comment might have made it a bit misunderstandable, I am sorry about that. However, I was not talking about complying with Floodgate's License (I think you thought I was talking about that that), instead, I was talking about complying with FastLogin's License (https://github.com/games647/FastLogin/blob/main/LICENSE). I've brought up Floodgate as an example on how they are dealing with License headers, because their software is released under the same license that FastLogin uses.

Tl;dr: FastLogin is released under the MIT license which requires copyright headers. Nothing related to other libraries.
Please correct me if I am wrong.

games647:
Argh okay. I read multiple posts that this isn't necessary MIT. I would interpret `copy of the software` to include a license into the binary (here: jar) and not in the source code. Furthermore, we are not distributing any singular file. However to be on the safe side, I could add it. It's no big deal.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
```diff
@@ -0,0 +1,103 @@
...
+                    player.getName());
+            
+            // kicking must be synchronous
+            // https://www.spigotmc.org/threads/asynchronous-player-kick-problem.168580/
+            Bukkit.getScheduler().runTask(plugin, new Runnable() {
+                public void run() {
+                    player.kickPlayer("This name is allready in use by a Premium Java Player");
+                }
+            });
+
```
games647, line 44: Missing return

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+        if (!isValidConfigValue(allowNameConflict)) {
+            plugin.getLog().error(
```
games647, line 47: Kick the player, otherwise they would time out.

games647: Furthermore it makes sense to check the configuration on plugin startup first. Otherwise we would do it more often than necessary.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Override
+    public void run() {
+        plugin.getLog().info(
+                "Player {} is connecting through Geyser Floodgate.",
+                player.getName());
+        String allowNameConflict = plugin.getCore().getConfig().getString("allowFloodgateNameConflict");
+        // check if the Bedrock player is linked to a Java account 
+        boolean isLinked = floodgatePlayer.getLinkedPlayer() != null;
+        if (allowNameConflict.equalsIgnoreCase("linked") && !isLinked) {
```
games647, line 32: Shouldn't we also check if the java username is the same?

games647 (changes requested):

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
Smart123s: No, because you can link BE and JE accounts with different usernames.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
Smart123s: I've moved it to onEnable()

games647:
>Decide if players with the same name should be auto logged in from different platforms (they have different UUIDs)

What about the offline UUID is that the same with the Floodgate UUID?

Smart123s:
> > Decide if players with the same name should be auto logged in from different platforms (they have different UUIDs)
> 
> What about the offline UUID is that the same with the Floodgate UUID?


## Conclusion (tl;dr)

I've done some test's and Floodgate seems to give it's players a unique UUID

## Test results

### Configuration

| Config key | Value |
| ------------- | ------------- |
| nameChangeCheck  | true  |
| autoRegister  | false  |
| autoLogin  | true  |
| everything floodgate related  | true  |

Changing the value of `premiumUUID` did not have an effect on the results.

### UUID table

| Player*  | UUID
| ------------- | ------------- |
| Cracked JE  | 763d1638-1980-3070-a96b-5bf4dc7a35d4  |
| Floodgate  | 00000000-0000-0000-0009-01fb95bfcd08  |
| Premium JE <br>_as reported by server logs_ <br>premiumUUID=`false`  | 763d1638-1980-3070-a96b-5bf4dc7a35d4  |
| Premium JE <br>_as reported by server logs_ <br>premiumUUID=`true`  | 4028\*\*\*\*-\*\*\*\*-\*\*\*\*-\*\*\*\*-\*\*\*\*\*\*\*\*5945  |
| Real Premium UUID <br>_as reported by Mojang API_  | 4028\*\*\*\*-\*\*\*\*-\*\*\*\*-\*\*\*\*-\*\*\*\*\*\*\*\*5945 |

_*All players are using the same name_

### Login order
1. Cracked JE
2. Floodgate
3. Cracked JE
4. Premium JE

games647:
Thanks that clears it up. I think we should document this behavior and implies a different permission and inventory if server owners choose to keep linked accounts.

Smart123s:
I'll do the documentation ofc. Also, these tests were done using non-linked accounts. If they are linked, than the JE's UUID _should_ be used. (I didn't test this one yet, will edit this post once I do that)
Also, with AuthMe (and possibly other plugins), the password is the same for every player with the same name.

RedSmarty:
Hi, i tried compiling your fork but it still removes the floodgate prefix, any fix for this as of yet?

Smart123s:
> Hi, i tried compiling your fork but it still removes the floodgate prefix, any fix for this as of yet?

No, nothing yet. I'd really like to tell you to view to https://github.com/games647/FastLogin/issues/493 but there's no usable information there either.
And to be honest, I haven't néven looked into this issue yet, but I'll definitely do it in the near future.
One workaround could be to make FastLogin inject the prefixes instead of Floodgate, but it'd be really hacky. 
I think that the root of the problem might be related to an event being cancelled by FastLogin, but I am not sure. As I've said earlier, I haven't looked into this.

RedSmarty:
Thanks for the fast response. I guess i will just stick to bungeecord for now.

Smart123s:
### Introduction
Okay, so at this point I want to do a little bit of heads-up about this Pull Request, where it is, where it's going, what are the plans for the future, and what I have done lately.

### Where did it start?
My initial plan with this PR was to implement Floodgate support for Bukkit without even touching BungeeCord and/or Core, so I wouldn't break Bungee. It's worth noting that I haven't even tried the Bungee version of FastLogin back then. Well, a lot of things changed since then.

### Supporting BungeeCord
Yesterday, I took a look at https://github.com/games647/FastLogin/issues/498 and realized that the features I've been working on _(namely autoLogin, autoRegister and floodgateNameConflict)_ are not present in Bungee, but could be implemented in a similar way to it's Bukkit counterpart. In fact, the FloodgateApi is the same for all platforms.

#### Moving code to core
So it got me thinking that most of my code could be reused for a BungeeCord implementation of the above mentioned features. Today, I decided to move the `low-level name conflict checking` code from Bukkit to the core. This is kind of a first step. Since it's high-level part is not in the core yet, I've kept it disabled in Bungee (via a boolean set to false).

#### Adding Bungee features in this Pull Request
As I've mentioned earlier, this PR was supposed to be Bukkit only (look at the title of the PR), however I think that at this point it would be reasonable to expand it, and add configuration options to BungeeCord too.

**Question 1:** Should I do it in this PR, or should I keep this Bukkit only, and open another PR for BungeeCord?

### Things that are not going to be done
I think that a proper fix for https://github.com/games647/FastLogin/issues/493 is out of scope for this PR. I've looked at it several times and it seems to be an upstream issue _somewhere_. A workaround could be to make FastLogin assign the prefixes instead of Floodgate. I haven't tried it, so I don't know how complicated would it be to achieve that. However, a drawback of this approach could be that if the original bug goes away, then we'll have to deal with duplicate prefixes.

**Question 2:** Should I implement the workaround for Floodgate `name-prefix` when using FastLoign with ProtocolLib  Bukkit?

### To Do
There are two tasks in the original PR message that haven't been completed yet:
> Decide if players with the same name should be auto logged in from different platforms (they have different UUIDs)

There are two reasons why this is important. Firstly, Floodgate name prefixes are not obligatory and can be turned off. Secondly, https://github.com/games647/FastLogin/issues/493 is still present.
My plan for this is to add new option to autoRegister and autoLogin called `non-conflicting` which would only login/register players whose name doesn't conflict with a premium Java player's name. (Although after looking at the code for five minutes, I've realized that it **might** require more code modifications than what I've originally expected.)

> Check if the changes didn't break Floodgate's Xbox authentication. Floodgate should block offline Bedrock players from connecting.

That's actually a hard one, because if I sign out of my Microsoft account on my phone, the it won't let me connect to any server. My idea is to find a cracked version of the game which supports joining cracked servers too. Or if anyone has another idea on how to test this, please let me know. Or if someone could post actual test results, it'd be really awesome.

### Closing thoughts
@games647 could you please answer the two questions that I've asked above. I've also left comments on the changes you requested. Could you please check if those are OK now?

Thank you for reading this rather long comment. Have a nice day.

games647 (approved):

Smart123s:
> Keep it for Bukkit. You shouldn't blow up a pull request. PRs in reasonable atomic steps are much better. They are easier to manage and merge. You can always expand later. Think of it like a little bit bigger than commits. New features, bug fixes that already provide an improvement should directly be integrated into it. It makes it easier to test.

Ok, you are right. I'll add a warning to `onEnable()` about #493 and leave the rest of the code as is.
I'll try to find a version of Minecraft Bedrock Edition that lets me connect to servers WITHOUT having to sign in to Xbox live to check if Floodgate's authentication system still works. I'll leave another comment once I found the app & ran the tests. Or if you've done it already, then please stop me from wasting my time. :smile:

>> I think that a proper fix for #493 is out of scope for this PR. I've looked at it several times and it seems to be an upstream issue somewhere. A workaround could be to make FastLogin assign the prefixes instead of Floodgate. I haven't tried it, so I don't know how complicated would it be to achieve that. However, a drawback of this approach could be that if the original bug goes away, then we'll have to deal with duplicate prefixes.

> It could be related to faking the receiving of a LoginStartPacket. We intercept this packet and create a new one. This is where the username will be and could be the reason of overwriting it.

It looks like I forgot to mention (although I though that I've done it before, sorry about that) that I've opened a ticket https://github.com/GeyserMC/Floodgate/issues/143. In short, the prefixes break after **registering** an async `start` listener using ProtocolLib, without doing anything in the actual eventhandler.

Smart123s:
I've reverted the commit where I have moved some code to Core since that part turned out to be useless for BungeeCord, so I thought that it shouldn't be merged.
I've also added warning when starting Floodgate Bukkit with either Floodgate 1.0 or 2.0 and ProtocolLib.
After several hours wasted, I was unable to find a Bedrock Edition client that allows you to connect to servers without having to sign in to Xbox live, so I gave up on that.
I'm not planning more code changes for this Pull Request.

bungee/src/main/resources/bungee.yml
```diff
@@ -9,6 +9,7 @@ version: ${project.version}-${git.commit.id.abbrev}
 author: games647, https://github.com/games647/FastLogin/graphs/contributors
 
 softDepends:
+    - floodgate
```
Smart123s, line 12: Looks like I've added this in another commit, so it didn't get reverted.
I'll keep it here since it should have been added in another Pull Request anyways.

games647: Then please include it in the other commit. Commit should be atomic. Otherwise it gets chaotic if we cherry-pick or revert certain commits. You could use the rebase functionality to override an existing commit.

games647:
>It looks like I forgot to mention (although I though that I've done it before, sorry about that) that I've opened a ticket GeyserMC/Floodgate#143. In short, the prefixes break after registering an async start listener using ProtocolLib, without doing anything in the actual eventhandler.

I experienced that somewhere else too, but I don't remember it. There is the importance of naming commits clearly. 

Could also be a ProtocolLib issue in this case, depending how the listener actually differs from the synchronous solution.

bukkit/src/main/resources/plugin.yml
```diff
@@ -20,6 +20,7 @@ softdepend:
     - ProtocolLib
     # Premium variable
     - PlaceholderAPI
+    - floodgate
```
games647, line 23: The mentioned removal

games647 (changes requested):

bungee/src/main/resources/bungee.yml
Smart123s: I have rebased it as you have requested. And don't worry, you are not strict in any ways, the changes you've asked for so far were all totally reasonable.

bukkit/src/main/resources/plugin.yml
Smart123s: Done

games647 (approved):

Trinchezito:
>.< nothing so far, anyone found alternative plugin similar to fastlogin?

PR #504, opened 2021-04-05
Add SodionAuthBungee support

InkerBot:

### Summary of your change

Bump SodionAuth version

Add SodionAuth-Bungee support

### Related issue

### DIFF
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/SodionAuthHook.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/SodionAuthHook.java
@@ -32,14 +32,16 @@ import red.mohist.sodionauth.bukkit.implementation.BukkitPlayer;
 import red.mohist.sodionauth.core.SodionAuthApi;
 import red.mohist.sodionauth.core.exception.AuthenticatedException;
 
+import java.util.concurrent.ExecutionException;
+
 /**
  * GitHub: https://github.com/Mohist-Community/SodionAuth
  * <p>
- * Project page:
+ * Project page: https://gitea.e-loli.com/SodionAuth/SodionAuth
  * <p>
  * Bukkit: Unknown
  * <p>
- * Spigot: Unknown
+ * Spigot: https://www.spigotmc.org/resources/sodionauth.76944/
  */
 public class SodionAuthHook implements AuthPlugin<Player> {
 
@@ -67,6 +69,9 @@ public class SodionAuthHook implements AuthPlugin<Player> {
         } catch (UnsupportedOperationException e){
             plugin.getLog().warn("Currently SodionAuth is not accepting forceRegister, " +
                     "It may be caused by unsupported AuthBackend");
+            return false;
+        } catch (InterruptedException | ExecutionException e) {
+            e.printStackTrace();
             return false;
         }
     }
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/SodionAuthHook.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/SodionAuthHook.java
@@ -32,8 +32,6 @@ import red.mohist.sodionauth.bukkit.implementation.BukkitPlayer;
 import red.mohist.sodionauth.core.SodionAuthApi;
 import red.mohist.sodionauth.core.exception.AuthenticatedException;
 
-import java.util.concurrent.ExecutionException;
-
 /**
  * GitHub: https://github.com/Mohist-Community/SodionAuth
  * <p>
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/SodionAuthHook.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/SodionAuthHook.java
@@ -70,9 +70,6 @@ public class SodionAuthHook implements AuthPlugin<Player> {
             plugin.getLog().warn("Currently SodionAuth is not accepting forceRegister, " +
                     "It may be caused by unsupported AuthBackend");
             return false;
-        } catch (InterruptedException | ExecutionException e) {
-            e.printStackTrace();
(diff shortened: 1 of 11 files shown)


bungee/src/main/java/com/github/games647/fastlogin/bungee/hook/SodionAuthHook.java
```diff
@@ -0,0 +1,81 @@
...
+        } catch (UnsupportedOperationException e){
+            plugin.getLog().warn("Currently SodionAuth is not accepting forceRegister, " +
+                    "It may be caused by unsupported AuthBackend");
+            return false;
+        } catch (InterruptedException | ExecutionException e) {
+            e.printStackTrace();
+            return false;
+        }
+    }
+    public boolean isRegistered(String playerName) {
```
games647, line 78: Missing Override annotation and an empty line

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/SodionAuthHook.java
```diff
@@ -68,6 +70,9 @@ public boolean forceRegister(Player player, String password) {
             plugin.getLog().warn("Currently SodionAuth is not accepting forceRegister, " +
                     "It may be caused by unsupported AuthBackend");
             return false;
+        } catch (InterruptedException | ExecutionException e) {
+            e.printStackTrace();
```
games647, line 74: Could you make use of the logger with a short message what we were trying here? At least we should give the users some hints about the error. Something like "Unexpected error during ..." should already do it.

bungee/src/main/java/com/github/games647/fastlogin/bungee/FastLoginBungee.java
```diff
@@ -121,6 +122,13 @@ private void registerHook() {
             BungeeCordAuthenticatorHook hook = new BungeeCordAuthenticatorHook(BungeeCordAuthenticatorBungee, logger);
             core.setAuthPluginHook(hook);
         }
+        Plugin SodionAuth = getProxy().getPluginManager().getPlugin("SodionAuth");
```
games647, line 125: Could you make this a loop? Explicitly listening the hook process seems like a lot of repetition.

games647 (changes requested):
Looks good so far. Only some minor things. Thanks

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/SodionAuthHook.java
InkerBot: I don't think api should throw such exception. I will remove it from SodionAuth

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/DelayedAuthHook.java
```diff
@@ -94,13 +94,13 @@ private boolean registerHooks() {
 
     private AuthPlugin<Player> getAuthHook() {
         try {
-            @SuppressWarnings("unchecked")
             List<Class<? extends AuthPlugin<Player>>> hooks = Arrays.asList(AuthMeHook.class,
                     CrazyLoginHook.class, LogItHook.class, LoginSecurityHook.class,
                     SodionAuthHook.class, UltraAuthHook.class, xAuthHook.class);
 
             for (Class<? extends AuthPlugin<Player>> clazz : hooks) {
-                String pluginName = clazz.getSimpleName().replace("Hook", "");
```
games647, line 103: I prefer to keep it explicit or make it `...length() - "Hook".length`. Otherwise it's a magic number, where the origin is not really obvious.

bungee/src/main/java/com/github/games647/fastlogin/bungee/FastLoginBungee.java
```diff
@@ -110,16 +116,22 @@ public void onDisable() {
...
+            List<Class<? extends AuthPlugin<ProxiedPlayer>>> hooks = Arrays.asList(
+                    BungeeAuthHook.class, BungeeCordAuthenticatorBungeeHook.class, SodionAuthHook.class);
+
+            for (Class<? extends AuthPlugin<ProxiedPlayer>> clazz : hooks) {
+                String pluginName = clazz.getSimpleName();
+                pluginName = pluginName.substring(0, pluginName.length() - 4);
+                //uses only member classes which uses AuthPlugin interface (skip interfaces)
+                Plugin plugin = getProxy().getPluginManager().getPlugin(pluginName);
+                if (plugin != null) {
+                    core.setAuthPluginHook(
```
games647, line 129: A logging statement would be useful to inform the users about the plugin name, we hooked into.

bungee/src/main/java/com/github/games647/fastlogin/bungee/FastLoginBungee.java
```diff
@@ -110,16 +116,22 @@ public void onDisable() {
...
-            logger.info("Try to hook into BungeeCordAuthenticatorBungee...");
-            BungeeCordAuthenticatorHook hook = new BungeeCordAuthenticatorHook(BungeeCordAuthenticatorBungee, logger);
-            core.setAuthPluginHook(hook);
+        try {
+            List<Class<? extends AuthPlugin<ProxiedPlayer>>> hooks = Arrays.asList(
+                    BungeeAuthHook.class, BungeeCordAuthenticatorBungeeHook.class, SodionAuthHook.class);
+
+            for (Class<? extends AuthPlugin<ProxiedPlayer>> clazz : hooks) {
+                String pluginName = clazz.getSimpleName();
+                pluginName = pluginName.substring(0, pluginName.length() - 4);
```
games647, line 125: Same thing about the length as above.

games647 (changes requested):

games647:
Return after the first one. I don't think it's necessary to look further after the first one was found.

games647 (approved):
Perfect. Thank you very much.

PR #529, opened 2021-05-14
Use SodionCore download Sqlite driver in bungee

InkerBot:

### Summary of your change
Auto download SqliteDriver when SodionCore installed.

### Related issue

### DIFF
--- a/bungee/pom.xml
+++ b/bungee/pom.xml
@@ -141,6 +141,15 @@
             <scope>provided</scope>
         </dependency>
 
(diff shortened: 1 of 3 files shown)


games647:
Sorry, but I don't think we should add a complete dependency for this minor functionality. In my opinion such a library should be shaded directly into the plugin. What the BungeeCord is currently implementing looks promising however. https://github.com/SpigotMC/BungeeCord/commit/6a039de8db3a9dc78fe884e5d150db24df1933bb

This would allow dynamic loading of depencies at runtime.

PR #532, opened 2021-05-22
Share Floodgate name conflict check between Protocol Plugins

Smart123s:
### Summary of your change

- Added a shared class for Floodgate name conflict checking that can be used by both ProtocolLib and ProtocolSupport
- Polished the method of checking conflicting names
  Now a Floodgate related method checks the name and kicks the player if needed.
  Previously, Bedrock players with conflicting names were treated as Premium Java players. This way, they got an 'Invalid Session' some time during the Java auth process.
- Made ProtocolSupport check for Floodgate name conflicts
- Fixed allowFloodgateNameConflict = `linked`
  Previously, this kicked all non linked players.

### Related issue
ProtocolSupport now checks for Floodgate name conflicts (no open issue)


Smart123s:
Whit the latest rebase I have:
- removed some `e.printStackTrace();`
- added back the last empty line to ProtocolSupportListener.java

Smart123s:
I think I haven't mentioned it yet, but this PR is ready for review.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/floodgate/FloodgateHook.java
```diff
@@ -0,0 +1,89 @@
...
+import org.geysermc.floodgate.api.player.FloodgatePlayer;
+
+import com.github.games647.craftapi.model.Profile;
+import com.github.games647.craftapi.resolver.RateLimitException;
+import com.github.games647.fastlogin.bukkit.FastLoginBukkit;
+import com.github.games647.fastlogin.core.shared.LoginSource;
+
+public class FloodgateHook {
+
+    FastLoginBukkit plugin;
```
games647, line 17: No access modifier

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/NameCheckTask.java
```diff
@@ -70,13 +69,18 @@ public NameCheckTask(FastLoginBukkit plugin, PacketEvent packetEvent, Random ran
...
         try {
-            // check if the player is connecting through Geyser
-            if (!plugin.getCore().getConfig().get("allowFloodgateNameConflict").toString().equalsIgnoreCase("false")
-                    && getFloodgatePlayer(username) != null) {
-                plugin.getLog().info("Skipping name conflict checking for player {}", username);
-                return;
+            ProtocolLibLoginSource source = new ProtocolLibLoginSource(packetEvent, player, random, publicKey);
+
+            //check if the player is connecting through Floodgate
+            FloodgateHook floodgateHook = new FloodgateHook(plugin);
```
games647, line 75: Initiating this doesn't make much sense, because it doesn't have any player specific state. It would be better to have a single instance in the protocol classes and then pass the service through to the tasks.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/floodgate/FloodgateHook.java
```diff
@@ -0,0 +1,89 @@
...
+        // check if the Bedrock player is linked to a Java account 
+        boolean isLinked = floodgatePlayer.getLinkedPlayer() != null;
+
+        if (allowConflict.equals("false")
+                || allowConflict.equals("linked") && !isLinked) {
+
+            // check for conflicting Premium Java name
+            Optional<Profile> premiumUUID = Optional.empty();
+            try {
+                premiumUUID = plugin.getCore().getResolver().findProfile(username);
```
games647, line 44: This makes an additional UUID request. It's very rate-limited, so can't we reuse the existing check from the login request?

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/floodgate/FloodgateHook.java
```diff
@@ -0,0 +1,89 @@
...
+            } catch (IOException | RateLimitException e) {
+                plugin.getLog().error(
+                        "Could not check wether Floodgate Player {}'s name conflits a premium Java player's name.",
+                        username);
+                try {
+                    source.kick("Could not cehck if your name conflits an existing Java Premium Player's name");
```
games647, line 50: Typo

games647 (changes requested):

Smart123s:
## About rate limiting

> It's very rate-limited

I know that and I also want to make sure that there won't be more than one request sent.

> so can't we reuse the existing check from the login request

Are you referring to this?
https://github.com/games647/FastLogin/blob/e8bb9fd30a82cf07e4358d4a55185032979b8e07/core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java#L80

I assume that you do.
### Firstly
That won't be called if a player is connecting through Floodgate.

https://github.com/games647/FastLogin/blob/757d0ef99140f04b3e69c3f8c01bd05b6ef15320/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocolsupport/ProtocolSupportListener.java#L88-L93

https://github.com/games647/FastLogin/blob/757d0ef99140f04b3e69c3f8c01bd05b6ef15320/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/NameCheckTask.java#L80-L85

The checks done in [`super.onLogin(username, source);`](https://github.com/games647/FastLogin/blob/757d0ef99140f04b3e69c3f8c01bd05b6ef15320/core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java#L48) are all related to checking Java players if they are premium or use a premium name or etc. For Floodgate players, running these check would be a waste of power, because Floodgate players will always be detected as cracked. The only useful part of that code for Floodgate players is the name checking, so I've implemented it in a separate class. Running only the necessary checks saves server resources.

Since [`super.onLogin(username, source);`](https://github.com/games647/FastLogin/blob/757d0ef99140f04b3e69c3f8c01bd05b6ef15320/core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java#L48) is skipped, I had to use this line for a second time:
https://github.com/games647/FastLogin/blob/757d0ef99140f04b3e69c3f8c01bd05b6ef15320/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/floodgate/FloodgateHook.java#L44

Additionally, I have a rather complicated explanation on why there won't be multiple requests sent at https://github.com/games647/FastLogin/pull/533#issuecomment-846439811

### Secondly
The results of these requests should be cached anyways.

https://github.com/games647/CraftAPI/blob/f9c4a799016aa547366e14aee7a75f3a3c011160/src/main/java/com/github/games647/craftapi/resolver/MojangResolver.java#L155-L159

So storing the values a second time would just make the code less readable, without any benefit.

### Closing
These are just my thoughts and the way I **THINK** the code. You are still the owner of the code, and you still know things better than me, so please correct me, if I'm wrong somewhere.
Or if you think that creating a unified method for the API call is the way to go, then I'll do it of course.

Also, I have completed that other changes you have requested & rebased the PR.

Smart123s:
On the rate limiting (again).

I have done some work in a separate branch: https://github.com/Smart123s/FastLogin/tree/fg-ncshareapi

I have tried to kinda merge those API calls. Now there are still two of them, but there is a very clear 
```java
if(floodgate) {
    check api & do other stuff;
    return;
}
```
which makes it a lot easier to understand what it does.
You can see that code in https://github.com/Smart123s/FastLogin/commit/744264d36c4773adf1c38b0b181ba1c20e65d868

However, I ran to some problems with ProtocolSupport, see https://github.com/Smart123s/FastLogin/commit/45ece6939ddf9fd151bc7eb95517b0dd0240b758 (read commit message as well)

I don't have any more time today, so I'd like your opinion @games647 on this one. Is this a better approach? If so, I'll continue to work on this code, fix it up, etc. If not, I'll abandon it in it's current hacked-together form. 
If this is good, do you have any idea on how to get around the ProtocolSupport issue? I didn't have it with ProtocolLib.
My only idea is to put every bit of code that requires FloodgatePlayer references into a separate calss and then call them via a void. This say, ProtocolSupport won't cry, but the code will be ugly. (I've tried this, but deleted that code since, I'll rewrite it, if it'll be needed.)
Yes, I know that using unavailable classes as a return type is not the best practice, but I don't really have another choice. I try to check if Floodgate is available in every function which might have unsafe calls. However, ProtocolSupport doesn't even let me do that.

PS: The code mentioned above has some design flows, but it show the general approach on the API calls. I don't want to spend too much time with that code, until I don't know if it's actually a reliable approach.

games647:
>However, I ran to some problems with ProtocolSupport, see Smart123s@45ece69 (read commit message as well)

We could do a wrapper around it like a subclass. What I'd like to do in the future (if I find some time), is to seperate the optional features (i.e. protocollib, protocolsupport, bungeecord) into their own modules in order to reduce conflicts if they are not present like in this case. However in this case it's a little more complicated, because we can have a protocol plugin and floodgate at the same time.

When the logic is the same in BungeeCord, then your mentioned commit seems to better.

games647 (approved):
Points resolved

Smart123s:
I've pushed some additional commits. Now there are no `Object`s  used instead of `FloodgatePlayer`s.
Instead of overriding a `getFloodgatePlayer()` in every subclass of `JoinManagement`, there's now one unified method in `core/FloodghateHook`. The only downside to this approach is that we can't use UUIDs on BungeeCord check for Floodgate players. Bukkit only has usernames at the packet level, and now, Bungee will also use usernames to run the early Floodgate related checks.

Also, the current changes will have no effect on BungeeCord (for now), because the previous Floodgate Integration is still in place, and it stops the code before it could reach this PR's code.
https://github.com/games647/FastLogin/blob/d1109ac75e2f5362587bef8e016749918622ad08/bungee/src/main/java/com/github/games647/fastlogin/bungee/listener/ConnectListener.java#L125-L127
I have left it there intentionally, so that the existing integration of Bungee won't get messed up (for now). I plan to work on Bungee soon.

The commit history is now... _how do i say it?_ - interesting. I could squash some of the commits if you also think that it's necessary.

Apart from the awkward structure of the commits, the code is ready for a _hopefully_ final review.

games647 (approved):
LGTM

PR #533, opened 2021-05-22
Add 'no-conflict' option to some Floodgate config entries

Smart123s:
### Summary of your change
Added a `no-conflict` option to `autoRegisterFloodgate` and `autoLoginFloodgate` in config.yml
If this value is used, the player will be only logged in / registered if no Premium Java MC account exists with their name.

### Related issue
Could be useful for https://github.com/games647/FastLogin/issues/493

### DIFF
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
@@ -25,10 +25,15 @@
  */
 package com.github.games647.fastlogin.bukkit.task;
 
+import java.io.IOException;
+import java.util.Optional;
+
 import org.bukkit.Bukkit;
 import org.bukkit.entity.Player;
 import org.geysermc.floodgate.api.player.FloodgatePlayer;
 
+import com.github.games647.craftapi.model.Profile;
+import com.github.games647.craftapi.resolver.RateLimitException;
 import com.github.games647.fastlogin.bukkit.BukkitLoginSession;
 import com.github.games647.fastlogin.bukkit.FastLoginBukkit;
 import com.github.games647.fastlogin.core.StoredProfile;
@@ -58,7 +63,8 @@ public class FloodgateAuthTask implements Runnable {
         AuthPlugin<Player> authPlugin = plugin.getCore().getAuthPluginHook();
 
         String autoLoginFloodgate = plugin.getCore().getConfig().get("autoLoginFloodgate").toString().toLowerCase();
-        boolean autoRegisterFloodgate = plugin.getCore().getConfig().getBoolean("autoRegisterFloodgate");
+        String autoRegisterFloodgate = plugin.getCore().getConfig().get("autoRegisterFloodgate").toString().toLowerCase();
+        String allowNameConflict = plugin.getCore().getConfig().get("allowFloodgateNameConflict").toString().toLowerCase();
         
         boolean isRegistered;
         try {
@@ -70,7 +76,27 @@ public class FloodgateAuthTask implements Runnable {
             return;
         }
         
-        if (!isRegistered && !autoRegisterFloodgate) {
+        //decide if checks should be made for conflicting Java player names
+        if (autoLoginFloodgate.equals("no-conflict")
+                || !isRegistered && autoRegisterFloodgate.equals("no-conflict")) {
+            // check for conflicting Premium Java name
+            Optional<Profile> premiumUUID = Optional.empty();
+            try {
+                premiumUUID = plugin.getCore().getResolver().findProfile(player.getName());
+            } catch (IOException | RateLimitException e) {
+                plugin.getLog().error(
+                        "Could not check wether Floodgate Player {}'s name conflits a premium Java player's name.",
+                        player.getName());
+                return;
+            }
+
+            //stop execution if player's name is conflicting
+            if (premiumUUID.isPresent()) {
+                return;
+            }
+        }
+
+        if (!isRegistered && autoRegisterFloodgate.equals("false")) {
             plugin.getLog().info(
                     "Auto registration is disabled for Floodgate players in config.yml");
             return;
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
@@ -76,7 +76,10 @@ public class FloodgateAuthTask implements Runnable {
         }
 
         //decide if checks should be made for conflicting Java player names
-        if (!isLinked &&
+        if (!isLinked //linked players have the same name as their Java profile
+                // if allowNameConflict is 'false' or 'linked' and the player had a conflicting
+                // name, than they would have been kicked in FloodgateHook#checkNameConflict
+                && allowNameConflict.equals("true") &&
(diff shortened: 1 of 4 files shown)


Smart123s:
I've encountered some bugs with this one, don't merge it yet.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
```diff
@@ -59,21 +66,42 @@ public void run() {
...
+
+        //decide if checks should be made for conflicting Java player names
+        if (autoLoginFloodgate.equals("no-conflict")
+                || !isRegistered && autoRegisterFloodgate.equals("no-conflict")) {
+            // check for conflicting Premium Java name
+            Optional<Profile> premiumUUID = Optional.empty();
+            try {
+                premiumUUID = plugin.getCore().getResolver().findProfile(username);
+            } catch (IOException | RateLimitException e) {
+                e.printStackTrace();
```
games647, line 78: The logger is responsible for stacktrace formatting. Otherwise it would end up in stdio. Just use the existing logging message.

games647: Rate limiting is a big issue. Try to reduce as much as possible. In this case, couldn't we use the result from the login if we made a check there?

core/src/main/resources/config.yml
```diff
@@ -223,8 +227,13 @@ autoLoginFloodgate: false
...
 allowFloodgateNameConflict: false
 
-# This enables auto registering every player connecting through Floodgate.
-# autoLoginFloodgate must be 'true' for this to work
+# Automatically register players connecting through Floodgate.
+# autoLoginFloodgate must be available for the player to use this
+# Possible values:
+#   false: Disables auto registering for every player connecting through Floodgate
+#   true: Enables auto registering for every player connecting through Floodgate
+#   no-conflict: Bedrock players will only be automatically registered if no Premium Java MC account exists
```
games647, line 235: Please state that no-conflicts performs a possible rate-limiting check against Mojang.

games647 (changes requested):

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java
Smart123s: `e.printStackTrace();` was auto generated by Eclipse and I forgot to remove it. Will do it.

> couldn't we use the result from the login

If you are refering to 
https://github.com/games647/FastLogin/blob/main/core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java#L80 then no, because that whole function will be skipped if a player is connecting from Floodgate. After all, Floodgate players will never be true premium Java players.

Smart123s:
Sorry, I screwed up the rebase, commits should be fine now.

core/src/main/resources/config.yml
Smart123s: Rephrased the sentence to include that & rebased the commit.

Smart123s:
> Rate limiting is a big issue. Try to reduce as much as possible. In this case, couldn't we use the result from the login if we made a check there?
As far as I can tell, it can be called only once, but correct me if I'm wrong.

There are currently 3 `plugin.getCore().getResolver().findProfile(username);` in the code:

### First API call code

The first one is in upstream:
https://github.com/games647/FastLogin/blob/119b9cb00061f1a226478dec7bd6a50eb6b7737f/core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java#L80
However, this code will only be reached if the player is NOT using Floodgate, after https://github.com/games647/FastLogin/pull/532 gets merged (maybe merge that one first?)
https://github.com/games647/FastLogin/blob/aaddc249302e0f35d426044afa4e9e06cb99114e/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/NameCheckTask.java#L78-L83

### Second API call code

The second one is currently only in #532 https://github.com/games647/FastLogin/blob/aaddc249302e0f35d426044afa4e9e06cb99114e/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/floodgate/FloodgateHook.java#L39
This one can be reached based on the value of `allowFloodgateNameConflict` which can be `true`, `false` or `linked`
If `true`, this part of the code will be skipped
`false` and `linked` will be discussed in the third part

### Third API call code

The third one is in this Pull Request
https://github.com/games647/FastLogin/blob/157c9466db844cec01a0af2329dee0b7adf18bd5/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java#L76
If `allowFloodgateNameConflict` = `false` then every conflicting player will be kicked in part 2, so this part won't be reachable.
_and the lines after the highlighted code snippet, which I won't copy here_
https://github.com/games647/FastLogin/blob/aaddc249302e0f35d426044afa4e9e06cb99114e/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/hook/floodgate/FloodgateHook.java#L34

If `allowFloodgateNameConflict` = `linked` then they'll 100% have a conflicting name, so there's no need to check for that.
See https://github.com/games647/FastLogin/commit/2c2415b63accd519f6d296bc19036f4bd91bd566

### When double request coccus
If `allowFloodgateNameConflict` = `false` or `linked` and either `autoLoginFloodgate` nor `autoRegisterFloodgate` is set to `no-conflict` and a player with a non-conflicting name connects that there will be a double check. I'll fix that in the next commit.

### Why bother with having name conflict checks on 3 different places?
Why not just check it when the player joins?
Because in the first & second phase we don't know if checking will be neccessary.
For example, if `allowFloodgateNameConflict` = `true` and neither `autoLoginFloodgate` nor `autoRegisterFloodgate` is set to `no-conflict` that name checking is not needed.

### Did I miss something?
If I did, please let me know.

Smart123s:
I've decided not to mark this pull request as "Ready for review" until #532 get reviewed & merged.
While this PR could work as a standalone patch without any fatal problems (or I'd better say without any _new_ problems), it'd be reasonable to merge after #532.
For example, a double API call would be instantly fixed without having to modify a single line of code this PR.

I've opened separate PRs, because these features can work standalone, without one another. However, while writing this, I had the previous PR in mind.
Why didn't I wait for it to get finalized before starting this PR? Because today I had plenty of time to code and I wanted to use it.

Smart123s:
I have rebased the PR on branch [main](https://github.com/games647/FastLogin/tree/main)@https://github.com/games647/FastLogin/commit/01c9b55d80c93e0df4ad9050519995f4d56db759

Smart123s:
## API calls

After https://github.com/games647/FastLogin/pull/533/commits/fb8780d0067864447bdb5c8b1e09ac63b7a7b649 the name checking API call will be only made if `allowNameConflict` is set to `true` in config.yml.

And if it's set to `true`, then no (name conflict checking) API calls were made while the player was connecting (I'm talking about the packet level), because if Floodgate players are allowed to connect with conflicting names, than we don't have to check if their names are conflicting while they are connecting.
https://github.com/games647/FastLogin/blob/01c9b55d80c93e0df4ad9050519995f4d56db759/core/src/main/java/com/github/games647/fastlogin/core/hooks/FloodgateHook.java#L60-L66
Do I have to document this in the source code via comments too?

games647 (approved):
LGTM

PR #534, opened 2021-05-22
Warn in config: linked players conflict their Java name

Smart123s:
### Summary of your change
Add the following warning to `allowFloodgateNameConflict` in config.yml
> Linked players have the same name as their Java profile, so the Bedrock player will always conflict#     their own Java account's name. Therefore, setting this to false will prevent any linked player from joining.

I've decided to keep the default value `false` because in it's current state, Floodgate integration might have unknown bugs, and `false` is the safest value. This should be changed in the future. Or should I change it now?


games647:
>Or should I change it now?

No it's fine as it is.

games647 (approved):

PR #543, opened 2021-06-13
Get rid of several warning messages

Smart123s:
### Summary of your change
Remove unused variables and imports.
Every warning fix is in a separate commit (except for the imports), so it'll be easier to revert one of them, if they'd be useful in the future. If one (or more) of them is needed, then I'll revert it.

### Related issue
The IDE will throw less warnings. It really annoyed me, but I didn't want to hide these commits in another PR, so I've made a separate PR.


games647 (approved):

games647:
LGTM

PR #558, opened 2021-06-20
Add Floodgate config options for BungeeCord

Smart123s:
### Summary of your change
- Moved most of [Bukkit/FloodgateAuthTask](https://github.com/games647/FastLogin/blob/578028719f0bf8f7ed2066d2746216434d0e64c0/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java) to core
- Removed previous Floodgate integration
  This has automatically enabled `allowFloodgateNameConflict`, since all of it's code is in core.
  The previous integration used a `return;` statement before this code could have gotten executed.
- Copied the remaining part of [Bukkit/FloodgateAuthTask](https://github.com/games647/FastLogin/blob/578028719f0bf8f7ed2066d2746216434d0e64c0/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java) to [Bungee/FloodgateAuthTask](https://github.com/Smart123s/FastLogin/blob/0b307a95a341a186b8a8f20b0b3012740fd2fe30/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/task/FloodgateAuthTask.java) and modified it to work with BungeeCord
- Added a fake online mode option to [Bungee/ForceLoginTask](https://github.com/Smart123s/FastLogin/blob/0b307a95a341a186b8a8f20b0b3012740fd2fe30/bungee/src/main/java/com/github/games647/fastlogin/bungee/task/ForceLoginTask.java), so that Floodgate players can auto login.

- Updated config.yml

### Related issue
No specific issue was opened, but I remember some comments requesting this in several unrelated and now closed issues.

### Dropped support for Floodgate 1.0
- The Bukkit implementation has never supported Floodgate 1.0
- Floodgate 2.0 is now officially released.
- From my experience, updating Floodgate is just a simple drop-in replacement.
- Also, Floodgate 1.0 is incompatible with Waterfall 1.16 after build 419 and BungeeCord after build 1567.

For these reasons, I think that keeping support for a legacy version of Floodgate is unnecessary, so I've dropped it to make development easier.


core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
```diff
@@ -0,0 +1,131 @@
...
+            isRegistered = authPlugin.isRegistered(username);
+        } catch (Exception e) {
+            core.getPlugin().getLog().error(
+                    "An error has occured while checking if player {} is registered",
+                    username);
+            return;
+        }
+
+        //decide if checks should be made for conflicting Java player names
+        if (!isLinked //linked players have the same name as their Java profile
```
games647, line 86: I realized now this check is too complicated to read. Maybe you should extract it into a method or multiple program flows.

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
```diff
@@ -0,0 +1,131 @@
...
+            return;
+        }
+
+        // logging in from bedrock for a second time threw an error with UUID
+        profile = core.getStorage().loadProfile(username);
+        if (profile == null) {
+            profile = new StoredProfile(getUUID(player), username, true, getAddress(player).toString());
+        }
+
+        performLogin = true;
```
games647, line 123: This can also be implemented by calling an abstract method, that is implemented by subclasses. Might be a better solution than relying on a status variable.

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
```diff
@@ -0,0 +1,131 @@
...
+            }
+
+            //stop execution if player's name is conflicting
+            if (premiumUUID.isPresent()) {
+                return;
+            }
+        }
+
+        if (!isRegistered && autoRegisterFloodgate.equals("false")) {
+            core.getPlugin().getLog().info(
```
games647, line 112: This could be identified as SPAM. Maybe you could add the player name, so it's addressable.

games647 (changes requested):

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
Smart123s: I actually wanted to remove this weeks ago, but forgot to do it.

games647 (approved):

PR #525, opened 2021-05-08
Fix Floodgate 2.0 on BungeeCord

Smart123s:
### Summary of your change
In floodgate 2.0 the API instance was changed from `FloodgateAPI` to `FloodgateApi` which caused the API calls not to work.
A version checking has been added to determine if floodgate 1 or 2 is used and use the proper API.

This is just a hotfix and does not implement the features from https://github.com/games647/FastLogin/pull/494. I might do that in a future PR, but I think that existing features should be fixed first.

### Related issue
Fixes https://github.com/games647/FastLogin/issues/498

PR #559, opened 2021-06-20
Add 'linked' option to 'autoRegisterFloodgate'

Smart123s:
### Summary of your change
There is a new option available in `config.yml` for `autoRegisterFloodgate`:
https://github.com/games647/FastLogin/blob/2b7800e138ebc5b12c7095abef8a77d02a715536/core/src/main/resources/config.yml#L237

This option makes sense with Floodgate 2.0's [Global Link API](https://github.com/GeyserMC/Floodgate/wiki/Features#what-is-the-global-api)

### Related issue
None

### DIFF
--- a/core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
+++ b/core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
@@ -105,7 +105,7 @@ public abstract class FloodgateManagement<P extends C, C, L extends LoginSession
             }
         }
 
-        if (!isRegistered && autoRegisterFloodgate.equals("false")) {
+        if (!isRegistered && !isAutoRegisterAllowed()) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
```diff
@@ -120,6 +120,15 @@ public void run() {
 
     }
 
+    /**
+     * Decude if the player can be auto registered.
+     * The config option 'non-conflicting' is ignored by this function.
+     * @return true if the Player can be registered automatically
+     */
+    private boolean isAutoRegisterNotAllowed() {
```
games647, line 128: Checking for the not allowed seems to be a bit complicated. I think it's better to have `!isAllowed...` or what do you think?

games647 (dismissed):

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
Smart123s: It's a lot better that way.

games647 (approved):
LGTM.

BTW:
Thank you very much for all the effort you put into this. I really appreciate it.

Smart123s:
> BTW:
> Thank you very much for all the effort you put into this. I really appreciate it.

@games647 Thank you for reviewing and merging my work to your project, even after the stupid mistakes I've made, especially with https://github.com/games647/FastLogin/pull/525 and https://github.com/games647/FastLogin/pull/528. And honestly, you've done a fantastic job with this plugin. The idea of "reimplementing" Mojang authentication is amazing, and you also provide good support for your plugin. You are kind and helpful to people, even when they ask trivial questions that have already been answered before. A lot of people wouldn't do that. I honestly think that you've created a brilliant plugin, and I am really happy that I can be a part of the development, even if I can only contribute a small part to this wonderful project.

PR #606, opened 2021-08-26
Don't crash if no storage is loaded

Smart123s:

### Summary of your change
If the pluign is running on Bukkit, and it's connected to Bungee
then core.getStorage() will be null.
If that's the case, players will be logged in via a plugin messages.
### Related issue
https://github.com/games647/FastLogin/pull/602#issuecomment-905740940


games647:
Here in the `PlayerJoinEvent`. If there is no login session, then the bungee task wasn't run yet.

https://github.com/games647/FastLogin/blob/172efafc2bfe11ff2f6425f8b40be32918057c1d/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/ConnectionListener.java#L92

Smart123s:
I get it now, thanks. Although it seems like https://github.com/games647/FastLogin/pull/606#issuecomment-907177147 was also fixed by https://github.com/games647/FastLogin/pull/608.

games647 (approved):

Issue #603, opened 2021-08-22
Java Premium login is requested for Floodgate players on Bungee


Smart123s:
### What happened?

The Bukkit part of the server is irrelevant, because the Player gets kicked before reaching it.

This happens randomly (around 50% of the time), if the player is not yet in the Floodgate.getPlayers() list when doing the name conflict checking:
https://github.com/games647/FastLogin/blob/01c9b55d80c93e0df4ad9050519995f4d56db759/core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java#L61

If this line returns `null`, it means that he/she is a Java Player, and they'll be treated as one. Because my Bedrock username is the same as my Java username, I'm treated as a Premium player. Because of that, FastLogin will try to log in the Floodgate players as if they were a Premium Java players, which causes the errors that I've attached in the logs below.

I have experimented quite a bit with the code, and found out the following:

- The `username` variable in...
https://github.com/games647/FastLogin/blob/01c9b55d80c93e0df4ad9050519995f4d56db759/core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java#L53
...sometimes contains the Floodgate prefix, sometimes it doesn't.

- `connection.getUniqueId()` is `null` at the PreLogin event, although it should have been initialized by Floodgate.
  - Why should it have a value? 
    -  [Floodgate's PreLogin event priority](https://github.com/GeyserMC/Floodgate/blob/61a6ea99bfe2ae17946f9f65aa339447f80dbb95/bungee/src/main/java/org/geysermc/floodgate/listener/BungeeListener.java#L79-L80) is `LOWEST` while [FastLogin's](https://github.com/games647/FastLogin/blob/1f4757669c282833506c26866abfbec6355e51db/bungee/src/main/java/com/github/games647/fastlogin/bungee/listener/ConnectListener.java#L105-L106) is unset (=`NORMAL`).
    - This was in the docs of the Floodgate 1.0 implementation of FastLogin
https://github.com/games647/FastLogin/blob/d6a02aa5385ef7eb6c8b2f2169dfae1a351ef5b8/bungee/src/main/java/com/github/games647/fastlogin/bungee/listener/ConnectListener.java#L211-L213
[...]

PR #608, opened 2021-08-28
Delay Floodgate name checks on BungeeCord

Smart123s:

### Summary of your change
The `FloodgateApi` returns `null` at `PreLoginEvent` so all Floodgate related checks have to be done at `LoginEvent`.

### Related issue
Fixes https://github.com/games647/FastLogin/issues/603

### Why Draft?

- [ ] No one else has reported the issue, so there's no way to verify if it also affects other users.
- [ ] This is a fast workaround, so code quality is not the best. Please notify me about the necessary changes.
   I especially don't like how I had to create a new class for [BungeeFloodgateLoginSource](https://github.com/games647/FastLogin/blob/a46a5fb6116e3798689af53ae2ca75af7b1719ae/bungee/src/main/java/com/github/games647/fastlogin/bungee/BungeeFloodgateLoginSource.java), which is only used by 
https://github.com/games647/FastLogin/blob/a46a5fb6116e3798689af53ae2ca75af7b1719ae/bungee/src/main/java/com/github/games647/fastlogin/bungee/listener/ConnectListener.java#L159-L160
- [ ] Because of
https://github.com/games647/FastLogin/blob/a46a5fb6116e3798689af53ae2ca75af7b1719ae/bungee/src/main/java/com/github/games647/fastlogin/bungee/listener/ConnectListener.java#L132
- [ ] The skipped Floodgate checking code is still in core, it's just made unreachable via a `return` statement.
   _It was easier to make it stay there for now, because on Bukkit, there are two Protocol plugins, and both of them have to run the Floodgate checks._

### DIFF
--- a/bungee/pom.xml
+++ b/bungee/pom.xml
@@ -141,6 +141,14 @@
             <scope>provided</scope>
         </dependency>
 
+        <!-- Bedrock player bridge for low level checks -->
+        <dependency>
+            <groupId>org.geysermc</groupId>
+            <artifactId>connector</artifactId>
+            <version>1.2.1-SNAPSHOT</version>
+            <scope>provided</scope>
+        </dependency>
+
         <!--Login plugin-->
         <dependency>
             <groupId>me.vik1395</groupId>
--- a/bungee/src/main/java/com/github/games647/fastlogin/bungee/BungeeFloodgateLoginSource.java
+++ b/bungee/src/main/java/com/github/games647/fastlogin/bungee/BungeeFloodgateLoginSource.java
@@ -1,0 +1,79 @@
+/*
+ * SPDX-License-Identifier: MIT
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2021 <Your name and contributors>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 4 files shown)


EnriqueGF:
> ### Summary of your change
> The `FloodgateApi` returns `null` at `PreLoginEvent` so all Floodgate related checks have to be done at `LoginEvent`.
> 
> ### Related issue
> Fixes #603
> 
> ### Why Draft?
> * [ ]  No one else has reported the issue, so there's no way to verify if it also affects other users.
> * [ ]  This is a fast workaround, so code quality is not the best. Please notify me about the necessary changes.
>   I especially don't like how I had to create a new class for [BungeeFloodgateLoginSource](https://github.com/games647/FastLogin/blob/a46a5fb6116e3798689af53ae2ca75af7b1719ae/bungee/src/main/java/com/github/games647/fastlogin/bungee/BungeeFloodgateLoginSource.java), which is only used by
>   https://github.com/games647/FastLogin/blob/a46a5fb6116e3798689af53ae2ca75af7b1719ae/bungee/src/main/java/com/github/games647/fastlogin/bungee/listener/ConnectListener.java#L159-L160
> * [ ]  Because of
>   https://github.com/games647/FastLogin/blob/a46a5fb6116e3798689af53ae2ca75af7b1719ae/bungee/src/main/java/com/github/games647/fastlogin/bungee/listener/ConnectListener.java#L132
> * [ ]  The skipped Floodgate checking code is still in core, it's just made unreachable via a `return` statement.
>   _It was easier to make it stay there for now, because on Bukkit, there are two Protocol plugins, and both of them have to run the Floodgate checks._

Hello,

I'm experiencing the same issue, in my case a Floodgate player never is auto logged.

I tried implementing your changes, but is still not working for me.

Could you add me to Discord? EnriqueGF#0977

Thank's you for all.

Smart123s:
> Hello,
> 
> I'm experiencing the same issue, in my case a Floodgate player is never auto logged.
> 
> I tried several times implementing your changes, but is still not working for me.
> 
> I also tried to git clone your git repository with the fix branch of this but it takes no effect for me ingame.
> 
> Could you add me to Discord? EnriqueGF#977
> 
> Thank's you for all.

Ok, I'll add you, but I think you missed the last digit of your Discord Tag. It should be 4 digits long.

Smart123s:
Ok, I've added you.

EnriqueGF:
Ok I talked with @Smart123s about this.

I tried the changes but it doesn't work **only** if name checking is enabled.

The console throws something like "User connected through Floodgate, name conflicts will be checked later" but this never happens and the user is not logged in.

Issue #549, opened 2021-06-15
Velocity support


games647:
>doesn't give us an official and concrete answer about adding support to Velocity? I

I followed the development of Velocity closely till today. At the beginning there were too many breaking changes. A few months ago, I took another look. I got the idea to start working on it for Velocity 2.0. However after this [announcement](https://forums.velocitypowered.com/t/an-unfortunate-update-on-2-0-0-and-an-exciting-announcement/772), plans have shifted. 

So I'm now working towards implementing support for 1.1.X. However there are a few things I want to make to ensure future versions are easier to test, so changes can be made with less manual testing. The current plan is the following:

1. Add integration tests: automatic testing environment of authentication process
2. Separate major optional features into modules: ProtocolLib vs ProtocolSupport, etc
3. Add velocity support

Velocity support won't be that complicated, because the proxy logic is rather minimal in comparison to Spigot. Nevertheless adding a new platform always adds more maintenance work.

_Originally posted by @games647 in https://github.com/games647/FastLogin/issues/311#issuecomment-861442453_

games647:
FastLogin modules will depend on Velocity 3.0. It features an event system for asynchronous processing which seems to be very helpful for this project.

MAX-XiaoKui:
So what auth plugin will be supported?

games647:
Velocity only covers the proxy area. You don't need a proxy auth plugin. FastLogin will support bridged auth plugins from BungeeCord, but also the already existing in Spigot.

MAX-XiaoKui:
What if I have vanilla or fabric subservers?

games647:
This would require support for this platform too. However a proxy only could also work out.

kubo6472:
The SNAP workaround stopped working as of move to velocity 3.0 as the stable and official version. Logs: https://hastebin.com/zajasiniqi.properties

MAX-XiaoKui:
> The SNAP workaround stopped working as of move to velocity 3.0 as the stable and official version. Logs: https://hastebin.com/zajasiniqi.properties

FastLogin doesn't support Snap.
See #509

kubo6472:
> FastLogin doesn't support Snap.

Hi, thanks for the link. It worked for me on v1.1.x somehow. But now it does not anymore. Gotta wait for the official support then.

games647:
> > The SNAP workaround stopped working as of move to velocity 3.0 as the stable and official version. Logs: https://hastebin.com/zajasiniqi.properties
> 
[...]

PR #612, opened 2021-09-14
Velocity support

juanmuscaria:
### Summary of your change
This pull request adds support for Velocity.
It's currently based on the bungeecord version with all the logic replaced with Velocity-equivalent classes. It's my first time working with bungeecord and velocity so please correct any mistake I've made. I removed all floodgate/hooks related code because I'm unable to test bedrock connections. 
It also needs some cleanups, I made it in a rush due to needing to migrate to velocity asap. Tomorrow I'll cleanup the code and do more testings  together with implementing requested changes.
In my current testings no issues raised so far and it's working with an authme fork for velocity
image

### Related issue
Related to #549

### DIFF
--- a/velocity/pom.xml
+++ b/velocity/pom.xml
@@ -97,12 +97,6 @@
             <id>velocity</id>
             <url>https://nexus.velocitypowered.com/repository/maven-public/</url>
         </repository>
-
-        <!--For MySql Driver-->
-        <repository>
-            <id>Maven Central</id>
-            <url>https://repo1.maven.org/maven2/</url>
-        </repository>
     </repositories>
 
     <dependencies>
--- a/velocity/pom.xml
+++ b/velocity/pom.xml
@@ -71,14 +71,15 @@
                             <shadedPattern>fastlogin.hikari</shadedPattern>
                         </relocation>
                         <relocation>
-                            <pattern>org.slf4j</pattern>
-                            <shadedPattern>fastlogin.slf4j</shadedPattern>
-                        </relocation>
-                        <relocation>
                             <pattern>net.md_5.bungee.config</pattern>
                             <shadedPattern>fastlogin.config</shadedPattern>
                         </relocation>
                     </relocations>
+                    <artifactSet>
+                        <excludes>
+                            <exclude>org.slf4j:*</exclude>
+                        </excludes>
+                    </artifactSet>
                 </configuration>
                 <executions>
                     <execution>
--- a/velocity/pom.xml
+++ b/velocity/pom.xml
@@ -84,6 +84,7 @@
                         shading and relocating it causes logger injection to fail-->
                         <excludes>
                             <exclude>org.slf4j:*</exclude>
+                            <exclude>com.google.code.gson:gson</exclude>
                         </excludes>
                     </artifactSet>
                 </configuration>
--- a/velocity/pom.xml
+++ b/velocity/pom.xml
@@ -46,6 +46,19 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>templating-maven-plugin</artifactId>
+                <version>1.0.0</version>
+                <executions>
+                    <execution>
+                        <id>filter-src</id>
+                        <goals>
+                            <goal>filter-sources</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
                 <version>3.2.4</version>
--- a/velocity/pom.xml
+++ b/velocity/pom.xml
@@ -80,33 +80,15 @@
     </build>
 
     <repositories>
-
-        <repository>
-            <id>codemc-repo</id>
-            <url>https://repo.codemc.io/repository/maven-public/</url>
-        </repository>
-
-        <repository>
-            <id>opencollab-snapshot</id>
-            <url>https://repo.opencollab.dev/maven-snapshots/</url>
-        </repository>
-
-        <repository>
-            <id>spigotplugins-repo</id>
-            <url>https://maven.gamestrike.de/mvn/</url>
-        </repository>
-
         <repository>
             <id>velocity</id>
             <url>https://nexus.velocitypowered.com/repository/maven-public/</url>
         </repository>
 
+        <!--For MySql Driver-->
         <repository>
-            <id>jitpack.io</id>
-            <url>https://jitpack.io</url>
-            <snapshots>
-                <enabled>false</enabled>
-            </snapshots>
+            <id>Maven Central</id>
+            <url>https://repo1.maven.org/maven2/</url>
         </repository>
     </repositories>
 
@@ -126,11 +108,12 @@
             <scope>provided</scope>
         </dependency>
 
+        <!--Velocity does not ship any database driver-->
         <dependency>
-            <groupId>org.xerial</groupId>
-            <artifactId>sqlite-jdbc</artifactId>
-            <version>3.32.3</version>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
+            <version>8.0.26</version>
+            <scope>runtime</scope>
         </dependency>
-
     </dependencies>
 </project>
--- a/velocity/pom.xml
+++ b/velocity/pom.xml
@@ -80,6 +80,8 @@
                         </relocation>
                     </relocations>
                     <artifactSet>
+                        <!--org.slf4j is part of velocity api and runtime,
+                        shading and relocating it causes logger injection to fail-->
                         <excludes>
                             <exclude>org.slf4j:*</exclude>
                         </excludes>
--- a/velocity/pom.xml
+++ b/velocity/pom.xml
@@ -125,10 +125,9 @@
 
         <!--Velocity does not ship any database driver-->
         <dependency>
-            <groupId>mysql</groupId>
-            <artifactId>mysql-connector-java</artifactId>
-            <version>8.0.26</version>
-            <scope>runtime</scope>
+            <groupId>org.mariadb.jdbc</groupId>
+            <artifactId>mariadb-java-client</artifactId>
+            <version>2.7.4</version>
         </dependency>
     </dependencies>
 </project>
--- a/velocity/pom.xml
+++ b/velocity/pom.xml
@@ -1,0 +1,136 @@
+<!--
+
+    SPDX-License-Identifier: MIT
+
+    The MIT License (MIT)
+
+    Copyright (c) 2015-2021 <Your name and contributors>
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
+    in the Software without restriction, including without limitation the rights
+    to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+    copies of the Software, and to permit persons to whom the Software is
+    furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included in all
+    copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.github.games647</groupId>
+        <artifactId>fastlogin</artifactId>
+        <version>1.11-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <!--This have to be in lowercase because it's used by plugin.yml-->
+    <artifactId>fastlogin.velocity</artifactId>
+    <packaging>jar</packaging>
+
+    <!--Represents the main plugin-->
+    <name>FastLoginVelocity</name>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>3.2.4</version>
+                <configuration>
+                    <createDependencyReducedPom>false</createDependencyReducedPom>
+                    <shadedArtifactAttached>false</shadedArtifactAttached>
+                    <relocations>
+                        <relocation>
+                            <pattern>com.zaxxer.hikari</pattern>
+                            <shadedPattern>fastlogin.hikari</shadedPattern>
+                        </relocation>
+                        <relocation>
+                            <pattern>org.slf4j</pattern>
+                            <shadedPattern>fastlogin.slf4j</shadedPattern>
+                        </relocation>
+                        <relocation>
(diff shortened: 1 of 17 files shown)


velocity/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+            <id>codemc-repo</id>
+            <url>https://repo.codemc.io/repository/maven-public/</url>
+        </repository>
+
+        <repository>
+            <id>opencollab-snapshot</id>
+            <url>https://repo.opencollab.dev/maven-snapshots/</url>
+        </repository>
+
+        <repository>
```
games647, line 94: Please remove not needed repositories. It reduces build time, because it scans those repositories for dependencies too.

So only velocity-repo should be needed.

velocity/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+
+        <!--Velocity api-->
+        <dependency>
+            <groupId>com.velocitypowered</groupId>
+            <artifactId>velocity-api</artifactId>
+            <version>3.0.1</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
```
games647, line 129: SQLite doesn't provide the same performance than MySQL/MariaDB instance. Therefore we don't ship it for proxy instances that likely have a higher user base.

velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
```diff
@@ -0,0 +1,125 @@
...
+import java.nio.file.Path;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentMap;
+
+import com.velocitypowered.api.proxy.messages.MinecraftChannelIdentifier;
+import com.velocitypowered.api.proxy.server.RegisteredServer;
+import net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer;
+import org.slf4j.Logger;
+
+@Plugin(id = "fastlogin")
```
games647, line 39: Please add the other meta data too. The version number could be quite useful for example in server logs and more. You could let them generated from the Maven project data like I did in ChangeSkin.

velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
```diff
@@ -0,0 +1,125 @@
...
+public class FastLoginVelocity implements PlatformPlugin<CommandSource> {
+    private final ProxyServer server;
+    private final Path dataDirectory;
+    private final Logger logger;
+    private FastLoginCore<Player, CommandSource, FastLoginVelocity> core;
+    private final ConcurrentMap<InetSocketAddress, VelocityLoginSession> session = new MapMaker().weakKeys().makeMap();
+    private AsyncScheduler scheduler;
+
+    @Inject
+    public FastLoginVelocity(ProxyServer server, java.util.logging.Logger logger, @DataDirectory Path dataDirectory) {
```
games647, line 49: `java.util.Logger` is only used for backwards compatibility. If possible, we can use SL4F directly, not need to create a wrapper.

velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
```diff
@@ -0,0 +1,125 @@
...
+        server.getChannelRegistrar().register(MinecraftChannelIdentifier.create(getName(), ChangePremiumMessage.CHANGE_CHANNEL));
+        server.getChannelRegistrar().register(MinecraftChannelIdentifier.create(getName(), SuccessMessage.SUCCESS_CHANNEL));
+        server.getEventManager().register(this, new ConnectListener(this, core.getRateLimiter()));
+        server.getEventManager().register(this, new PluginMessageListener(this));
+    }
+
+    @Override
+    public String getName() {
+        //FIXME: some dynamic way to get it?
+        return "fastlogin";
```
games647, line 75: A constant like mentioned above for the annotation.

velocity/src/main/java/com/github/games647/fastlogin/velocity/task/ForceLoginTask.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Override
+    public String getName(Player player) {
+        return player.getUsername();
+    }
+
+    @Override
+    public boolean isOnline(Player player) {
+        //FIXME: is this right?
+        return core.getPlugin().getProxy().getPlayer(player.getUsername()).isPresent();
```
games647, line 134: This could end up wrong. If there is another login request is coming in, this would evaluate to true even though it's not the same connection or even person.

velocity/src/main/java/com/github/games647/fastlogin/velocity/task/ForceLoginTask.java
```diff
@@ -0,0 +1,141 @@
...
+    public FastLoginAutoLoginEvent callFastLoginAutoLoginEvent(LoginSession session, StoredProfile profile) {
+        //FIXME: Am I doing it right?
+        VelocityFastLoginAutoLoginEvent event = new VelocityFastLoginAutoLoginEvent(session, profile);
+        try {
+             return core.getPlugin().getProxy().getEventManager().fire(event).get();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt(); // Set the interrupt flag again
+            return event;
+        } catch (ExecutionException e) {
+            e.printStackTrace();
```
games647, line 101: Always use the logger for the exceptions, to ensure it has the correct output.

games647 (changes requested):
Thanks for working on it. I took a quick peak at it. I'll take a deeper look later when I have some spare time.

The proxy ID system might not be necessary, because Velocity doesn't seem to suffer from malicious player plugin messages. This could be the case for players directly connecting to the server. However there is still the risk that a player could send a force login plugin message. If the proxy login crashed or is not enabled (including by mistake), the message could get forwarded to the Spigot server, so we need to verify if it's still necessary for this case.

Another thing is the asynchronous event processing. This is a main feature of velocity and is extremely important for anti bot protection. I didn't spend much time with it, but it seems to fix the design flaw BungeeCord has. I think it's important that we integrate it right from the beginning.

https://velocitypowered.com/wiki/developers/event-api/#:~:text=Handling%20events%20asynchronously

velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
juanmuscaria: Velocity only provides you with `java.util.logging.Logger` for logging, so I did the conversion like it's made on the bukkit and bungee version. I avoided messing around anything outside the velocity implementation to prevent making any unwanted change.

velocity/src/main/java/com/github/games647/fastlogin/velocity/task/ForceLoginTask.java
juanmuscaria: I missed that there was an equivalent for `#isConnected()`, my bad.

juanmuscaria:
A few other things I'll have to look into will be testing each individual option and implement what I missed. I'll not be able to implement anything related to floodgate due to me being unable to test the implementation, but a velocity version of floodgate is available to integrate compatibility .

velocity/pom.xml
games647: Thanks, but then we should also relocate the package to prevent any version breakages.

velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
games647: I thought Velocity provides out of the box support for it, because the guide also indicates it: https://velocitypowered.com/wiki/developers/api-basics/

games647:
>Adding a new configuration option for a "secret" could work as replacement for the proxy id, other option is using velocity's built-in proxy secret and generate an uuid from it. Velocity allows the legacy bungee way to connect to a server, and probably most people will still use it because of older server version it it's probably a good idea to have something similar to proxy id.

For this case, we still need it. Otherwise it could cause security issues with an invalid setup. In theory the current implementation of FastLogin is not really secure. A better solution would be the implementation of Velocity with a correct signature like HMAC, but that doesn't protect against replay attacks. However it seems to be necessary in our case.

For this PR, I think it's the best to port the existing implementation and save a random UUID in the plugins folder of the proxy. So we could use that as a persistent ID. Any breaking changes like above (signing changes), should be handled in a seperate PR.

>The listeners I could probably remove the usage of the async scheduler because the event is already async and makes use of continuation.

That's kind of what I meant. Asynchronous processing of the event, so multiple users could process the event in parallel. As far as I know that's what the continuation feature promises. However I noticed now that you are already integrated it for the relevant part. The other events don't need it.

velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
juanmuscaria: My bad, it indeed supports it, the error it was generating was not helpful but I finally tracked down to the issue being the relocation of org.slf4j.Logger on the shade plugin which caused the inject to fail.

velocity/pom.xml
juanmuscaria: I don't think it's possible to relocate an jdbc driver without also changing a few things in de driver itself, and seems like it's normal for a velocity plugin to ship the needed database driver. In case the multiple drivers are present only the first driver loaded will be used, so the end user can either overwrite the driver if needed my manually adding it to the classpath.

juanmuscaria:
I did some testings and I think the velocity version is already in pair with the bungee version and ready for use, the only thing that will need to be implemented by someone else is floodgate support together with hooks for velocity auth plugins since so far I don't know any auth plugin for velocity.

velocity/pom.xml
juanmuscaria: With the way FastLogin loads drivers it can be relocated, users can use `fastlogin.mysql.cj.jdbc.Driver` on the configuration file to load the mysql driver or use any other driver provided by other plugins or from the classpath itself,

games647:
Do you tried the MySQL driver change? You could be correct that it doesn't work.

juanmuscaria:
Using the following config
```
driver: 'fastlogin.mysql.cj.jdbc.Driver'
host: '127.0.0.1'
port: 3306
database: 'test'
username: 'test'
password: 'password'
```
Seems to work fine with the shaded mysql driver, I know the sqlite driver will have problems because it uses native, the plugin does fail to load if no driver is shaded

games647 (approved):

velocity/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+                            <pattern>net.md_5.bungee.config</pattern>
+                            <shadedPattern>fastlogin.config</shadedPattern>
+                        </relocation>
+                        <relocation>
+                            <pattern>com.mysql</pattern>
+                            <shadedPattern>fastlogin.mysql</shadedPattern>
+                        </relocation>
+                    </relocations>
+                    <artifactSet>
+                        <excludes>
```
games647, line 83: Please mentioned why we are excluding slf4j, like it's included in Velocity. It's better to be explicit for later times.

velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
```diff
@@ -0,0 +1,192 @@
...
+//TODO: Support for floodgate
+@Plugin(id = PomData.NAME, name = PomData.DISPLAY_NAME, description = PomData.DESCRIPTION, url = PomData.URL,
+        version = PomData.VERSION, authors = {"games647", "https://github.com/games647/FastLogin/graphs/contributors"})
+public class FastLoginVelocity implements PlatformPlugin<CommandSource> {
+
+    private final ProxyServer server;
+    private final Path dataDirectory;
+    private final Logger logger;
+    private final ConcurrentMap<InetSocketAddress, VelocityLoginSession> session = new MapMaker().weakKeys().makeMap();
+    private final String PROXY_ID_fILE = "proxyId.txt";
```
games647, line 73: Please make this a global constant (`static`). The compiler can then inline it, not really relevant for performance, but a good convention.

velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
```diff
@@ -0,0 +1,192 @@
...
+        if (Files.exists(idFile)) {
+            try {
+                List<String> lines = Files.readAllLines(idFile, StandardCharsets.UTF_8);
+                if (lines.isEmpty()) {
+                    shouldGenerate = true;
+                } else {
+                    proxyId = UUID.fromString(lines.get(0));
+                }
+            } catch (IOException e) {
+                e.printStackTrace();
```
games647, line 167: This should be removed.

velocity/src/main/java/com/github/games647/fastlogin/velocity/listener/ConnectListener.java
```diff
@@ -0,0 +1,130 @@
...
+            playerProfile.setId(verifiedUUID);
+            if (!plugin.getCore().getConfig().get("premiumUuid", true)) {
+                UUID offlineUUID = UUIDAdapter.generateOfflineId(playerProfile.getName());
+                event.setGameProfile(event.getGameProfile().withId(offlineUUID));
+                plugin.getLog().info("Overridden UUID from {} to {} (based of {}) on {}",
+                        verifiedUUID, offlineUUID, verifiedUsername, event.getConnection());
+            }
+
+            if (!plugin.getCore().getConfig().get("forwardSkin", true)) {
+                //FIXME: Do I need to remove *all* properties or only the skin related ones?
```
games647, line 98: Good point, only texture ones if that is possible. This should make it future proof.

velocity/src/main/java/com/github/games647/fastlogin/velocity/listener/ConnectListener.java
```diff
@@ -0,0 +1,130 @@
...
+        plugin.getProxy().getScheduler()
+                .buildTask(plugin, () -> plugin.getScheduler().runAsync(loginTask))
+                .delay(1L, TimeUnit.SECONDS) // Delay at least one second, otherwise the login command can be missed
+                .schedule();
+    }
+
+    @Subscribe
+    public void onDisconnect(DisconnectEvent disconnectEvent) {
+        Player player = disconnectEvent.getPlayer();
+        assert plugin.getSession().remove(player.getRemoteAddress()) != null;
```
games647, line 127: Correct me if I am wrong, but asserts doesn't work in without special runtime parameters. This is better solved in other programming languages ;)

games647 (changes requested):
Looks good, only some minor things. However I need some more time to check for security issues before merging it just to be sure.

games647:
> Using the following config
> 
> ```
> driver: 'fastlogin.mysql.cj.jdbc.Driver'
> host: '127.0.0.1'
> port: 3306
> database: 'test'
> username: 'test'
> password: 'password'
> ```
> 
> Seems to work fine with the shaded mysql driver, I know the sqlite driver will have problems because it uses native, the plugin does fail to load if no driver is shaded

Good to hear, thanks.

games647:
> Using the following config
> 
> ```
> driver: 'fastlogin.mysql.cj.jdbc.Driver'
> host: '127.0.0.1'
> port: 3306
> database: 'test'
> username: 'test'
> password: 'password'
> ```
> 
> Seems to work fine with the shaded mysql driver, I know the sqlite driver will have problems because it uses native, the plugin does fail to load if no driver is shaded

We should mention this maybe. Could you update the configuration file to include a note for the different driver name, but good to hear that it works.

velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
```diff
@@ -0,0 +1,192 @@
...
+    public FastLoginVelocity(ProxyServer server, Logger logger, @DataDirectory Path dataDirectory) {
+        this.server = server;
+        this.logger = logger;
+        this.dataDirectory = dataDirectory;
+    }
+
+    @Subscribe
+    public void onProxyInitialization(ProxyInitializeEvent event) {
+        scheduler = new AsyncScheduler(logger, getThreadFactory());
+        core = new FastLoginCore<>(this);
```
games647, line 89: Please add a close call for the core component. It will make sure that we are safely closing our resources, mainly the Hikari pool and our own scheduler that runs tasks velocity is not aware of.

games647 (changes requested):

velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
games647: Example:

https://github.com/games647/FastLogin/blob/ec7c421f839df286ceae52d1fb89d48556972e3a/bungee/src/main/java/com/github/games647/fastlogin/bungee/FastLoginBungee.java#L103

velocity/src/main/java/com/github/games647/fastlogin/velocity/listener/ConnectListener.java
juanmuscaria: That was a left over from when I was testing for a suitable object for the player session. I'll be removing it.

games647 (approved):

PR #614, opened 2021-09-15
Kick Java cracked user if it have the Floodgate Prefix

EnriqueGF:

### Summary of your change

With the actual code a cracked Java player can enter in the server with a username that contains a Floodgate prefix, like: ".Myusername" and he can register a new account, preventing a Floodgate user from registering.

The solution is to check the username of cracked users and kick the users who have a Floodgate prefix in the first character of the username.


games647:
Nice find, thanks.

games647 (approved):

Issue #672, opened 2021-12-04
Error between FastLogin and Floodgate (paper 1.18)


Danifoxe:
### What happened?

When starting Fastlogin, having the geyser and floodgate plugins, an error message appears and fastlogin is deactivated. 

Paper version git-Paper-36 (MC: 1.18) (Implementing API version 1.18-R0.1-SNAPSHOT) (Git: 76ee105)

### What did you expect?

Normal startup of both Fastlogin, Geyser and Floodgate. 

### Steps to reproduce

Have Fastlogin, Geyser and Floodgate on the same server paper 1.18. 

### Plugin list

Plugins (28): AuthMe, BuycraftX, CoreProtect, DiscordSRV, dynmap, Essentials, EssentialsAntiBuild, EssentialsChat, EssentialsSpawn, FastLogin, floodgate, Geyser-Spigot, GPRealEstate, GriefPrevention, GriefPreventionDynmap*, GriefPreventionFlags, GSit, LuckPerms, Parties, ProtocolLib, Quests, QuickShop, SkinsRestorer, TimeVote, Vault, Votifier, VotingPlugin, WorldEdit

### Configuration file

https://www.toptal.com/developers/hastebin/enaxuzocob.yaml

### Server log

```shell
https://www.toptal.com/developers/hastebin/ezixekuzew.md
```

### Plugin version

1.11-SNAPSHOT-95251b6

### Platform

Spigot

### Relevance

- [X] I tried the latest build
- [X] I checked for existing tickets -
If there are, please vote them with a thumps reaction and not create new ones

Smart123s:
The Geyser API was updated: https://github.com/GeyserMC/Geyser/wiki/Using-Geyser-or-Floodgate-as-a-dependency/_compare/7110301b7486b068eb685c4e4220b05cd18d80f5

I'll update FastLogin to use the new API as soon as I have some spare time. Support for the older API can be dropeed, because the Bedrock clients are always auto updated to the latest version, and they don't have a version selector.

PR #674, opened 2021-12-06
Update to Geyser 2.0 API (fix 1.18)

Smart123s:
### Summary of your change
Update to the Geyser 2.0 API.
It's not documented yet, but it's mentioned [here](https://github.com/GeyserMC/Geyser/wiki/Using-Geyser-or-Floodgate-as-a-dependency/_compare/7110301b7486b068eb685c4e4220b05cd18d80f5).

I have only done two test on this:
- I was able to auto log in with Floodgate
- The server starts without Geyser/Floodgate (and premium Java players can auto login)

### Related issue
Fixes https://github.com/games647/FastLogin/issues/672

### DIFF
--- a/bukkit/pom.xml
+++ b/bukkit/pom.xml
@@ -216,7 +216,7 @@
         <!-- Bedrock player bridge -->
         <dependency>
             <groupId>org.geysermc</groupId>
-            <artifactId>connector</artifactId>
+            <artifactId>core</artifactId>
             <version>${geyser.version}</version>
             <scope>provided</scope>
             <exclusions>
@@ -225,6 +225,14 @@
                     <artifactId>*</artifactId>
                 </exclusion>
             </exclusions>
+        </dependency>
+
+        <!-- We need the API, but it was excluded above -->
+        <dependency>
+            <groupId>org.geysermc</groupId>
+            <artifactId>geyser-api</artifactId>
+            <version>${geyser.version}</version>
+            <scope>provided</scope>
         </dependency>
 
         <!--Provide premium placeholders-->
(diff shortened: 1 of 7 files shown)


bukkit/pom.xml
```diff
@@ -216,7 +216,33 @@
...
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <!-- If you remove these, build will fail -->
+        <dependency>
+            <groupId>org.geysermc</groupId>
+            <artifactId>geyser-api</artifactId>
```
games647, line 233: If the dependency is necessary, it shouldn't be excluded above. This exclusion rule exists to to prevent pulling in the other dependencies like jackson.

games647 (changes requested):

bukkit/pom.xml
Smart123s: Ok, so how should this be addressed? Core must be included, because that contains the ˛`AuthType` class. However, core is dependent on jackson. I have force pushed some changes? Is it OK?

games647: >Core must be included, because that contains the ˛AuthType class.

We could use multiple exclusion rules, so that the core dependency is still included, but not the other ones.

Smart123s: But then every time Geyser adds a new dependency, we'll have to exclude it manually in FastLogin.

Anyways, I've looked into that and the list of excluded dependencies from Geyser would be longer (if we exclude one per line) than just adding `geyser-api` explicitly. [[There](https://github.com/GeyserMC/Geyser/blob/2953ea951321069a43b5b9195f960654b477ed73/core/pom.xml#L39-L254) are more than 200 lines of unneeded dependencies in Geyser's pom.xml].

games647: Good point

games647 (approved):

Issue #700, opened 2022-01-14
Differentiate between Floodgate and Java players


games647:
In the Floodgate database, players are identified by their names. If you set `autoLoginFloodgate` to `false`, Floodgate players will be treated as cracked. And as a result, their name will be set to cracked in the database.

*EDIT: so in other words, the database thinks that Floodgate and Java players are the same, if they have the same name.*

I'm planning to differentiate Floodgate and Java players in the database, but it'll take some coding, and I'll also have to do some kind of backwards compatibility for existing players.

_Originally posted by @Smart123s in https://github.com/games647/FastLogin/issues/696#issuecomment-1004973262_

games647:
I think this is really necessary to do, because FastLogin heavily depends on it and follows different really different program flow. Is there something I can help with? Maybe think of a database migration model? So we could do incremental changes.

Smart123s:
The idea is to add another column to the database, called `isFloodgate`, similarly to `isPremium`. Honestly, I didn't really look into the database code, and its API implementation, but I think that it can be done by just copy-pasting isPremium code. (I actually realized that it's not that easy, I'll talk about it later).

The db migration is a harder topic. Creating a utility that would migrate every player to the new format on the first start of the server after the plugin was updated would be impossible. Currently, there's no a universal way to detect Floodgate players. That's why we need the change after all. Currently, the code sets `isPremium` to true for players who should be auto logged in.

A better implementation would be to migrate players when they join the server for the first time (after the plugin update). `isFloodgate` would be `null` for players that haven't been migrated. So if it's `null`, the plugin would check if the player is using Floodgate, and update the database.

But.... _(this section only applies to servers that are not using a Floodgate prefix)_ The initial issue with the current implementation was that there were both Floodgate and Java players using the same name. And since they are different players, they'll need a different row in the db, But they use the same name. Making `isFloodgate` a key in the db could solve this. But it'd look awful, and it'd require lots of changes do the db API. And this problem (that I've described in the begging of paragraph) also makes migration harder. Let's assume we use the migrator that work on player joins. How do we decide if the old db entry is for a Floodgate or for a Java player? We simply can't, because of the problems I've outlined in the second paragraph.

[...]

Issue #696, opened 2022-01-04
Players are removed from premium list if they join with floodgate


YellowZaki:
### What happened?

Players are removed from premium list if they join with floodgate, when they had enabled /premium on java.

https://github.com/games647/FastLogin/issues/633 have the same problem but they have premiumUuid: true and all the comments are related to uuids. I have premiumUuid:false so I don't think there is a issue with uuids.

premiumUuid: false
autoLoginFloodgate: false
allowFloodgateNameConflict: true (I had to because players with premium nicks couldn't join on bedrock even if they hadn't used /premium)
autoRegisterFloodgate: false
floodgatePrefixWorkaround: true

### What did you expect?

I expect players with premium enabled, being on offline mode when they enter in bedrock, so they are requested the authme password.
And when they rejoin on java, join with online mode on.

### Steps to reproduce

1. Player joins in java and type /premium successfully.
2. Player quit and joins from Minecraft Bedrock (Floodgate) and have to type authme password
3. Player is removed from premium list (Premium == 0 on database) as soon as he enters through Bedrock.

### Plugin list

Authme, Floodgate

### Configuration file

https://www.toptal.com/developers/hastebin/yoyagoxudi.yaml

### Plugin version

1.11-SNAPSHOT-4a5516c

### Platform

BungeeCord

### Relevance

- [X] I tried the latest build
- [X] I checked for existing tickets -
If there are, please vote them with a thumps reaction and not create new ones

Smart123s:
In the Floodgate database, players are identified by their names. If you set `autoLoginFloodgate` to `false`, Floodgate players will be treated as cracked. And as a result, their name will be set to cracked in the database.

I'm planning to differentiate Floodgate and Java players in the database, but it'll take some coding, and I'll also have to do some kind of backwards compatibility for existing players.

Issue #633, opened 2021-10-18
Floodgate bugs and exploits


TDMidkwhatisxd:
### What happened?

I joined with bedrock using an auto-registered premium username, and the username was removed from /premium list.

### What did you expect?

The /premium username to keep its Premium features and auto login when using Java.

### Steps to reproduce

Use same bedrock username as the premium java username.

P.S. I had the nametag from floodgate to ""  i mean empty, after i changed to "_" it didnt work.

### Plugin version

1.11-SNAPSHOT-ef6f0fc

### Platform

Spigot

### Relevance

- [X] I tried the latest build
- [X] I checked for existing tickets -
If there are, please vote them with a thumps reaction and not create new ones

TDMidkwhatisxd:
You don't understand what I mean. 
1. Plugin glitched and didnt apply the prefixes, even tho floodgate show that the prefix is working in logs.
2. What i mean, is that fastlogin turn already registered premium users to offline.

Issue #689, opened 2021-12-29
Setting `floodgatePrefixWorkaround` to true gives "Your username contains illegal characters" to Bedrock Clients


some1chan:
### What happened?

Title more-or-less says it all. 

If you set `floodgatePrefixWorkaround` to true, you'll get the message "Your username contains illegal characters" upon connecting to the server. This message will go away if this variable was set to false, or the plugin is unloaded/uninstalled.

### Extra Details
* I tried manually merging [this pull request](https://github.com/games647/FastLogin/pull/681), and building a JAR from it. [You can find it here, if you're curious.](https://github.com/some1chan/FastLogin) I don't know what I'm doing at all, but it did compile. It didn't solve the issue though.
* I created a new issue, instead of commenting on #632. I figured the original issue was solved, but had an edge case worth documenting in more detail with the template. If not though, that's alright!

### What did you expect?

A Bedrock client is able to auto-register and login, with the Floodgate prefix.

### Steps to reproduce

1. Set `floodgatePrefixWorkaround` to true in the `config.yml` file
2. Connect to the server on a Bedrock client

### Plugin list

AuthMe, CoreProtect, Essentials, EssentialsChat, EssentialsProtect, EssentialsSpawn, FastLogin, floodgate, Geyser-Spigot, GeyserSkinManager, LuckPerms, PlugManX (PlugMan), ProtocolLib, SkinsRestorer, Vault, ViaVersion

### Configuration file

https://gist.github.com/some1chan/def3e8efcad45e0252abc17c3ba9b74a

### Server log

```shell
[02:19:08] [Network Listener - #2/INFO]: [Geyser-Spigot] /CENSORED:61259 tried to connect!
[02:19:09] [Network Listener - #2/INFO]: [Geyser-Spigot] Player connected with username DummyUsername
[02:19:09] [defaultEventLoopGroup-5-1/INFO]: [Geyser-Spigot] DummyUsername (logged in as: DummyUsername) has connected to the Java server
[02:19:10] [FastLogin Pool Thread #0/INFO]: [FastLogin] Handling player .DummyUsername
[02:19:10] [FastLogin Pool Thread #0/INFO]: [FastLogin] Floodgate Prefix detected on cracked player
[02:19:10] [Geyser Spigot connection thread-4-2/INFO]: [floodgate] Floodgate player logged in as .DummyUsername disconnected
[02:19:10] [Server thread/INFO]: /CENSORED:0 lost connection: Disconnected
[...]

PR #709, opened 2022-01-22
Differentiate Floodgate players in database

Smart123s:
### Summary of your change
Please read #700.
Added a new column to the database called `Floodgate`.
This will make the life of Bedrock and Java players with the same name a lot easier.

### Why Draft
I've just started working on this, and I wanted to share the code while I'm progressing with it.
***<ins>At this stage you have to delete the old database, or the server will crash when someone tries to join it.</ins>***

### Todo
- [x] Database migrator (so the old database won't have to be deleted)
- [x] Migrate every part of the code to be aware of the new `Floodgate` field in the database
  - [x] ~~Migrate `activate` plugin message~~*
  - [x] ~~Migrate `/cracked` and `/premium` commands (self)~~*
    - [x] Fix issue described at "Floodgate detection" (issue #689)
  - [x] ~~Migrate `/cracked` and `/premium` commands (others)~~*
- [x] Check what the `Premium` field in the db does for Floodgate players (so check for leftover code)
- [ ] Debug why `uuid` and `lastIp` are `null` for Floodgate players in the database ([code](https://github.com/games647/FastLogin/blob/8a01ddc231d6981be4ed755d3669727fc3bed4f5/core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java#L171-L172))

_*No longer needed, becuase usernames are now unique (https://github.com/games647/FastLogin/pull/709#issuecomment-1184623441)._

### To Test
*just notes for myself, so that I won't forget it*
- [x] Server without Floodgate
- [x] BungeeCord
- [x] Premium/Cracked commands
- [x] [AsyncToggleMessage](https://github.com/games647/FastLogin/blob/9656aadb353b0061b40f405cecab3481cd422738/bungee/src/main/java/com/github/games647/fastlogin/bungee/task/AsyncToggleMessage.java)
- [x] Auto login/register with same name Java and Bedrock accounts
    Result: only works on the platform where premium login was enabled

### Floodgate detection
https://github.com/games647/FastLogin/issues/689 needs to be fixed. (I'm not linking that issue to his PR on purpose, as it's not fixed yet.)
In short, if ProtocolLib prefix workaround is enabled, the following code wont work, after the prefix is added:
https://github.com/games647/FastLogin/blob/9b04ea5c89e0b99bc273d28a1d9b0483c0a1b903/core/src/main/java/com/github/games647/fastlogin/core/hooks/bedrock/FloodgateService.java#L105-L114

### Related issue
Fixes https://github.com/games647/FastLogin/issues/700 
Fixes https://github.com/games647/FastLogin/issues/696
Fixes https://github.com/games647/FastLogin/issues/633

### DIFF
--- a/core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
+++ b/core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
@@ -82,7 +82,19 @@ public abstract class FloodgateManagement<P extends C, C, L extends LoginSession
         profile = core.getStorage().loadProfile(username);
 
         if (profile.isSaved()) {
-            if (!profile.isFloodgateMigrated()) {
+            if (profile.isFloodgateMigrated()) {
+                if (profile.getFloodgate() == FloodgateState.TRUE && isLinked) {
+                    core.getPlugin().getLog()
+                            .info("Player {} is already stored by FastLogin as a non-linked Bedrock Edition player",
+                                    username);
+                    return;
+                } else if (profile.getFloodgate() == FloodgateState.FALSE && isLinked) {
+                    profile.setFloodgate(FloodgateState.LINKED);
+                    core.getPlugin().getLog().info(
+                            "Player {} will be changed from a Java player to a linked Floodgate player",
+                            username);
+                }
+            } else {
                 if (isLinked) {
                     profile.setFloodgate(FloodgateState.LINKED);
                     core.getPlugin().getLog().info(
@@ -93,16 +105,6 @@ public abstract class FloodgateManagement<P extends C, C, L extends LoginSession
                     core.getPlugin().getLog().info(
                             "Player {} will be migrated to the v2 database schema as a Floodgate user", username);
                 }
-            } else if (profile.getFloodgate() == FloodgateState.TRUE && isLinked) {
-                core.getPlugin().getLog()
-                        .info("Player {} is already stored by FastLogin as a non-linked Bedrock Edition player",
-                                username);
-                return;
-            } else if (profile.getFloodgate() == FloodgateState.FALSE && isLinked) {
-                profile.setFloodgate(FloodgateState.LINKED);
-                core.getPlugin().getLog().info(
-                        "Player {} will be changed from a Java player to a linked Floodgate player",
-                        username);
             }
         } else {
             if (isLinked) {
--- a/core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
+++ b/core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
@@ -80,6 +80,38 @@ public abstract class FloodgateManagement<P extends C, C, L extends LoginSession
         }
 
         profile = core.getStorage().loadProfile(username);
+
+        if (profile.isSaved()) {
+            if (!profile.isFloodgateMigrated()) {
+                if (isLinked) {
+                    profile.setFloodgate(FloodgateState.LINKED);
+                    core.getPlugin().getLog().info(
+                            "Player {} will be migrated to the v2 database schema as a linked Floodgate user",
+                            username);
+                } else {
+                    profile.setFloodgate(FloodgateState.TRUE);
+                    core.getPlugin().getLog().info(
+                            "Player {} will be migrated to the v2 database schema as a Floodgate user", username);
+                }
+            } else if (profile.getFloodgate() == FloodgateState.TRUE && isLinked) {
+                core.getPlugin().getLog()
+                        .info("Player {} is already stored by FastLogin as a non-linked Bedrock Edition player",
+                                username);
+                return;
+            } else if (profile.getFloodgate() == FloodgateState.FALSE && isLinked) {
+                profile.setFloodgate(FloodgateState.LINKED);
+                core.getPlugin().getLog().info(
+                        "Player {} will be changed from a Java player to a linked Floodgate player",
+                        username);
+            }
+        } else {
+            if (isLinked) {
+                profile.setFloodgate(FloodgateState.LINKED);
+            } else {
+                profile.setFloodgate(FloodgateState.TRUE);
+            }
+        }
+
         AuthPlugin<P> authPlugin = core.getAuthPluginHook();
 
         try {
@@ -119,13 +151,17 @@ public abstract class FloodgateManagement<P extends C, C, L extends LoginSession
             }
         }
 
+        // defer auto registration, if it's not enabled in the config
         if (!isRegistered && !isAutoAuthAllowed(autoRegisterFloodgate)) {
             return;
         }
 
-        //logging in from bedrock for a second time threw an error with UUID
-        if (profile == null) {
-            profile = new StoredProfile(getUUID(player), username, true, getAddress(player).toString());
+        // stop the auto login procedure, if the current connection's parameters don't match the one stored in our
+        // database
+        // ex. we stored a LINKED account, but the current connection is not linked
+        if ((profile.getFloodgate() == FloodgateState.LINKED && !isLinked)
+            || (profile.getFloodgate() == FloodgateState.TRUE && isLinked)) {
+            return;
         }
 
         //start Bukkit/Bungee specific tasks
--- a/core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java
+++ b/core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java
@@ -52,9 +52,8 @@ public abstract class JoinManagement<P extends C, C, S extends LoginSource> {
 
         //check if the player is connecting through Bedrock Edition
         if (bedrockService != null && bedrockService.isBedrockConnection(username)) {
-            //perform Bedrock specific checks
+            //perform Bedrock specific checks and skip Java checks if no longer needed
             if (bedrockService.performChecks(username, source)) {
-                //skip Java checks, since they are not needed
                 return;
             }
         }
@@ -66,14 +65,16 @@ public abstract class JoinManagement<P extends C, C, S extends LoginSource> {
             return;
         }
 
-        if (!profile.isFloodgateMigrated()) {
+        if (profile.isFloodgateMigrated()) {
+            if (profile.getFloodgate() == FloodgateState.TRUE) {
+                // migrated and enabled floodgate player, however the above bedrocks fails, so the current connection
+                // isn't premium
+                return;
+            }
+        } else {
             profile.setFloodgate(FloodgateState.FALSE);
             core.getPlugin().getLog().info(
                     "Player {} will be migrated to the v2 database schema as a JAVA user", username);
-        } else if (profile.getFloodgate() == FloodgateState.TRUE) {
-            core.getPlugin().getLog().info("Player {} is already stored by FastLogin as a Bedrock Edition player",
-                    username);
-            return;
         }
 
         callFastLoginPreLoginEvent(username, source, profile);
--- a/core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java
+++ b/core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java
@@ -49,17 +49,31 @@ public abstract class JoinManagement<P extends C, C, S extends LoginSource> {
 
     public void onLogin(String username, S source) {
         core.getPlugin().getLog().info("Handling player {}", username);
+
+        //check if the player is connecting through Bedrock Edition
+        if (bedrockService != null && bedrockService.isBedrockConnection(username)) {
+            //perform Bedrock specific checks
+            if (bedrockService.performChecks(username, source)) {
+                //skip Java checks, since they are not needed
+                return;
+            }
+        }
+
         StoredProfile profile = core.getStorage().loadProfile(username);
+
+        //can't be a premium Java player, if it's not saved in the database
         if (profile == null) {
             return;
         }
 
-        //check if the player is connecting through Bedrock Edition
-        if (bedrockService != null && bedrockService.isBedrockConnection(username)) {
-            //perform Bedrock specific checks and skip Java checks, if they are not needed
-            if (bedrockService.performChecks(username, source)) {
-                return;
-            }
+        if (!profile.isFloodgateMigrated()) {
+            profile.setFloodgate(FloodgateState.FALSE);
+            core.getPlugin().getLog().info(
+                    "Player {} will be migrated to the v2 database schema as a JAVA user", username);
+        } else if (profile.getFloodgate() == FloodgateState.TRUE) {
+            core.getPlugin().getLog().info("Player {} is already stored by FastLogin as a Bedrock Edition player",
+                    username);
+            return;
         }
 
         callFastLoginPreLoginEvent(username, source, profile);
@@ -139,6 +153,12 @@ public abstract class JoinManagement<P extends C, C, S extends LoginSource> {
         if (core.getConfig().get("nameChangeCheck", false)) {
             StoredProfile storedProfile = core.getStorage().loadProfile(profile.getId());
             if (storedProfile != null) {
+                if (storedProfile.getFloodgate() == FloodgateState.TRUE) {
+                    core.getPlugin().getLog()
+                            .info("Player {} is already stored by FastLogin as a Bedrock Edition player.", username);
+                    return false;
+                }
+
                 //uuid exists in the database
                 core.getPlugin().getLog().info("GameProfile {} changed it's username", profile);
 
--- a/core/src/main/java/com/github/games647/fastlogin/core/storage/SQLStorage.java
+++ b/core/src/main/java/com/github/games647/fastlogin/core/storage/SQLStorage.java
@@ -26,11 +26,14 @@
 package com.github.games647.fastlogin.core.storage;
 
 import com.github.games647.craftapi.UUIDAdapter;
+import com.github.games647.fastlogin.core.StoredProfile;
+import com.github.games647.fastlogin.core.shared.FloodgateState;
 import com.zaxxer.hikari.HikariConfig;
 import com.zaxxer.hikari.HikariDataSource;
 import org.slf4j.Logger;
 
 import java.sql.Connection;
+import java.sql.DatabaseMetaData;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -50,12 +53,14 @@ public abstract class SQLStorage implements AuthStorage {
             + "`UUID` CHAR(36), "
             + "`Name` VARCHAR(16) NOT NULL, "
             + "`Premium` BOOLEAN NOT NULL, "
-            + "`Floodgate` BOOLEAN NOT NULL, "
             + "`LastIp` VARCHAR(255) NOT NULL, "
             + "`LastLogin` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, "
             //the premium shouldn't steal the cracked account by changing the name
             + "UNIQUE (`Name`) "
             + ')';
+
+    protected static final String ADD_FLOODGATE_COLUMN_STMT = "ALTER TABLE `" + PREMIUM_TABLE
+            + "` ADD COLUMN `Floodgate` INTEGER(3)";
 
     protected static final String LOAD_BY_NAME = "SELECT * FROM `" + PREMIUM_TABLE
             + "` WHERE `Name`=? LIMIT 1";
@@ -85,11 +90,23 @@ public abstract class SQLStorage implements AuthStorage {
         // choose surrogate PK(ID), because UUID can be null for offline players
         // if UUID is always Premium UUID we would have to update offline player entries on insert
         // name cannot be PK, because it can be changed for premium players
-
         //todo: add unique uuid index usage
         try (Connection con = dataSource.getConnection();
-             Statement createStmt = con.createStatement()) {
-            createStmt.executeUpdate(CREATE_TABLE_STMT);
+             Statement stmt = con.createStatement()) {
+            stmt.executeUpdate(getCreateTableStmt());
+
+            // add Floodgate column
+            DatabaseMetaData md = con.getMetaData();
+            if (isColumnMissing(md, "Floodgate")) {
+                stmt.executeUpdate(ADD_FLOODGATE_COLUMN_STMT);
+            }
+
+        }
+    }
+
+    private boolean isColumnMissing(DatabaseMetaData metaData, String columnName) throws SQLException {
+        try (ResultSet rs = metaData.getColumns(null, null, PREMIUM_TABLE, columnName)) {
+            return !rs.next();
         }
     }
 
@@ -101,7 +118,8 @@ public abstract class SQLStorage implements AuthStorage {
             loadStmt.setString(1, name);
 
             try (ResultSet resultSet = loadStmt.executeQuery()) {
-                return parseResult(resultSet).orElseGet(() -> new StoredProfile(null, name, false, false, ""));
+                return parseResult(resultSet).orElseGet(() -> new StoredProfile(null, name, false,
+                        FloodgateState.FALSE, ""));
             }
         } catch (SQLException sqlEx) {
             log.error("Failed to query profile: {}", name, sqlEx);
@@ -134,9 +152,19 @@ public abstract class SQLStorage implements AuthStorage {
 
             String name = resultSet.getString("Name");
             boolean premium = resultSet.getBoolean("Premium");
+            int floodgateNum = resultSet.getInt("Floodgate");
+            FloodgateState floodgate;
+
+            // if the player wasn't migrated to the new database format
+            if (resultSet.wasNull()) {
+                floodgate = FloodgateState.NOT_MIGRATED;
+            } else {
+                floodgate = FloodgateState.fromInt(floodgateNum);
+            }
+
             String lastIp = resultSet.getString("LastIp");
             Instant lastLogin = resultSet.getTimestamp("LastLogin").toInstant();
-            return Optional.of(new StoredProfile(userId, uuid, name, premium, lastIp, lastLogin));
+            return Optional.of(new StoredProfile(userId, uuid, name, premium, floodgate, lastIp, lastLogin));
         }
 
         return Optional.empty();
@@ -154,9 +182,10 @@ public abstract class SQLStorage implements AuthStorage {
                         saveStmt.setString(1, uuid);
                         saveStmt.setString(2, playerProfile.getName());
                         saveStmt.setBoolean(3, playerProfile.isPremium());
-                        saveStmt.setString(4, playerProfile.getLastIp());
+                        saveStmt.setInt(4, playerProfile.getFloodgate().getValue());
+                        saveStmt.setString(5, playerProfile.getLastIp());
 
-                        saveStmt.setLong(5, playerProfile.getRowId());
+                        saveStmt.setLong(6, playerProfile.getRowId());
                         saveStmt.execute();
                     }
                 } else {
@@ -165,7 +194,9 @@ public abstract class SQLStorage implements AuthStorage {
 
                         saveStmt.setString(2, playerProfile.getName());
                         saveStmt.setBoolean(3, playerProfile.isPremium());
-                        saveStmt.setString(4, playerProfile.getLastIp());
+                        saveStmt.setBoolean(3, playerProfile.isPremium());
+                        saveStmt.setInt(4, playerProfile.getFloodgate().getValue());
+                        saveStmt.setString(5, playerProfile.getLastIp());
 
                         saveStmt.execute();
                         try (ResultSet generatedKeys = saveStmt.getGeneratedKeys()) {
@@ -181,6 +212,14 @@ public abstract class SQLStorage implements AuthStorage {
         } catch (SQLException ex) {
             log.error("Failed to save playerProfile {}", playerProfile, ex);
         }
+    }
+
+    /**
+     * SQLite has a slightly different syntax, so this will be overridden by SQLiteStorage
+     * @return An SQL Statement to create the `premium` table
+     */
+    protected String getCreateTableStmt() {
+        return CREATE_TABLE_STMT;
     }
 
     @Override
--- a/core/src/main/java/com/github/games647/fastlogin/core/storage/SQLStorage.java
+++ b/core/src/main/java/com/github/games647/fastlogin/core/storage/SQLStorage.java
@@ -50,19 +50,23 @@ public abstract class SQLStorage implements AuthStorage {
             + "`UUID` CHAR(36), "
             + "`Name` VARCHAR(16) NOT NULL, "
             + "`Premium` BOOLEAN NOT NULL, "
+            + "`Floodgate` BOOLEAN NOT NULL, "
             + "`LastIp` VARCHAR(255) NOT NULL, "
             + "`LastLogin` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, "
             //the premium shouldn't steal the cracked account by changing the name
             + "UNIQUE (`Name`) "
             + ')';
 
-    protected static final String LOAD_BY_NAME = "SELECT * FROM `" + PREMIUM_TABLE + "` WHERE `Name`=? LIMIT 1";
-    protected static final String LOAD_BY_UUID = "SELECT * FROM `" + PREMIUM_TABLE + "` WHERE `UUID`=? LIMIT 1";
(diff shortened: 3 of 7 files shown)


Smart123s:
@games647 I've got some questions.

As of now, (on branch `main`/https://github.com/games647/FastLogin/commit/9b04ea5c89e0b99bc273d28a1d9b0483c0a1b903), FastLogin allows Floodgate and Java players to be registered with the same name (without prefixes). However, the database contains

https://github.com/games647/FastLogin/blob/af2df500577db8a01c9cf664477fa11921eb5e97/core/src/main/java/com/github/games647/fastlogin/core/storage/SQLStorage.java#L59-L60

which forbids registering two players with the same name.

While I am strongly **against** registering multiple players with the same name, because it can cause confusion with some plugins (especially auth plugins), but there are folks out there already using this approach. For example, see https://github.com/games647/FastLogin/issues/696.

I have came up with the following possibilities as a solution:
- Deny joining the server, if the name has been registered (=saved in db) from a different platform (=Floodgate/Java)
  - This would be a bit forcing in my opinion, considering that Floodgate *allows* that (via disabling username prefixes)
- Allow auto login/registering only for the platform the player (name) joined first
  - With the approach, the player from second platform doesn't have to be saved in the db
- Remove the `UNIQUE` constraint from the db table, and allow same named registrations from multiple platforms (as it was before)
  - Are there any checks that should be introduced in other parts of the code, if the `UNIQUE` constraint is removed?

games647:
To be honest name changing is difficult topic. I made the `UNIQUE` decision earlier, because we try to find the player profile on login based on the player name. If there are two, who should we pick? Allowing name changes would mean the old name is longer up to date and should be removed, therefore unique.

Correct me if I'm wrong, but the referenced ticket is about the login with the same player name, that is the same account owner? The best option would be to allow both, because Floodgate authentication is identical to onlinemode verification. It only differs that FastLogin doesn't perform the validation. However, I noticed, this should only apply to linked profiles. Otherwise it should be forbidden for the second platform. 

Although it would be possible to have two accounts with the same player name, because as far as I know Floodgate uses different UUID. Nevertheless it could cause conflicts like you mentioned.

Smart123s:
Note for myslef: https://github.com/games647/FastLogin/pull/709/commits/87b21fd7d01f22062cf24c9887b53e045c327c61 (and it's parents) have merge conflicts resolved with https://github.com/games647/FastLogin/commit/6beaf194ce93ef0ae21f7e71286717f15ab54c18. This branch is currently based on https://github.com/games647/FastLogin/commit/62d7320a6e3f37c6b1a0ef63456c29856e80c9d6, as I couldn't strat up a server with anything newer.

Smart123s:
This pull request is more or less ready for review. I'll finish the remaining TODOs, run some test, and probably do a rebase.

Commits 1c63e70, 25e71e8, ae9772c, 24d552d, eab6ef9 could have been a separate pull request (I can still open a new one, if you'd like me to), but I thought that they fix an important issue (the floodgate prefix bug with ProtocolLib) that it should be fixed before the new database columns get added.

I might replace the [getBedrockPlayer()](https://github.com/games647/FastLogin/blob/40ac3733944d845cab2f07d473820c810ff2a4db/core/src/main/java/com/github/games647/fastlogin/core/hooks/bedrock/FloodgateService.java#L98-L114) function with the channel extraction method from https://github.com/games647/FastLogin/pull/709/commits/1c63e7037c219ac1beee5ae6e8715f01e366f313 and https://github.com/games647/FastLogin/pull/709/commits/ae9772c0b028c47367961eb000f3ab50e39cf940, but that could be done in a future PR. It'd also require me to figure out how to extract the player's channel in Bungee and in ProtocolSupport (and Velocity in the future), which would take some time.

For migrating legacy databases I came up with the idea to add a new integer column called `DBVersion`. If the database will have to be updated again in the future, we could simply increase the value of DBVersion. Is this approach OK?

<hr>

Do you have any idea on how to make this line shorter?

https://github.com/games647/FastLogin/blob/ae9772c0b028c47367961eb000f3ab50e39cf940/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/ProtocolLibListener.java#L295

This didn't help: https://stackoverflow.com/questions/41056517/breaking-a-long-url-to-several-line-in-javadoc

games647:
>For migrating legacy databases I came up with the idea to add a new integer column called DBVersion. If the database will have to be updated again in the future, we could simply increase the value of DBVersion. Is this approach OK?

Well there also the approach of doing [migrations](https://book.cakephp.org/migrations/3/en/index.html). This what I heard is normally used when doing schema changes. However I don't know if they fit here, because we are doing a decision during the login process and not on the data itself.

>Do you have any idea on how to make this line shorter?

Well you can split the `a`-tag, but that's it. 
```xml
<a href="">
lorem ipsum
</a>
```

Attribute `href` could also be on the next line, but that just looks strange.

Smart123s:
> Well there also the approach of doing [migrations](https://book.cakephp.org/migrations/3/en/index.html). This what I heard is normally used when doing schema changes.

The pluign you provided uses PHP, and I don't think that it'd fit for a java project. However, [Flyway](https://flywaydb.org/documentation/tutorials/java) could be used, which has a Java API.

> However I don't know if they fit here, because we are doing a decision during the login process and not on the data itself.

There are three approaches I could think of:

1. (current) Have a DBVersion column to check if the player was migrated or not.
 - tbh, I'm starting to think that this is just unnecessary bloat
2. Set Floodgate as a nullable boolean, and use that instead of a DBVersion column
 - and maybe make it nullable if a new database is created (so it's not an old one being migrated), although that would fragment the db structure, which could cause headaches in the future
3. Keep Floodgate as `not null` and tell everyone who updates the plugin that Floodgate players have to log in using their passwords, and type `/premium` to migrate themselves.
 - this would be the best idea in the long term
 - Floodgate support is still experimental, it's written 5 times in the config ([1](https://github.com/games647/FastLogin/blob/833177933a2e670b6ef6a49ee0c44d426672f432/core/src/main/resources/config.yml#L214-L215), [2](https://github.com/games647/FastLogin/blob/833177933a2e670b6ef6a49ee0c44d426672f432/core/src/main/resources/config.yml#L226-L227), [3](https://github.com/games647/FastLogin/blob/833177933a2e670b6ef6a49ee0c44d426672f432/core/src/main/resources/config.yml#L249-L250), [4](https://github.com/games647/FastLogin/blob/833177933a2e670b6ef6a49ee0c44d426672f432/core/src/main/resources/config.yml#L263-L264), [5](https://github.com/games647/FastLogin/blob/833177933a2e670b6ef6a49ee0c44d426672f432/core/src/main/resources/config.yml#L272)), so I could use it as an excuse to tell people that I don't care if a Floodgate player hasn't written down their automatically registered password. Although I'm pretty sure there would be a lot of new issues opened, and lots of people would kindly ask me to go to hell.

Which one should I go with? I'd actually prefer the second one. Regardless, I could use [Flyway](https://flywaydb.org/documentation/tutorials/java) to add the missing column(s) to the database. Would that plugin/library be okay?

games647:
I rather was talking about the concept of having change based migrations. So one table which is used to keep track of the applied migrations and then seperate files for each migration. The implementation can be very simple, because the files can fully be read and then applied. So we then have something like:

migrations/01_base_table.sql (the current table for existing setups)
migrations/02_add_floodgate_differentiation.sql
migrations/03_track_name_changes.sql (example)
migrations/04_drop_something.sql (example)

I don't think we need a library for that. We check the number of applied migrations and then apply the sql file(s) of which are missing.

What do you think about that? This might be a bit overkill, because it's often used in combinations with many contributors (then also with timestamps) and frequent scheme changes.

>and maybe make it nullable if a new database is created (so it's not an old one being migrated), although that would fragment the db structure, which could cause headaches in the future

I guess that's the best option to migrate the old data while using the new structure for new players. We could later drop the backwards compatibility by migrating all nullable values to false.

bukkit/pom.xml
```diff
@@ -376,7 +376,6 @@
             <groupId>io.netty</groupId>
             <artifactId>netty-transport</artifactId>
             <version>${nettyVersion}</version>
-            <scope>test</scope>
```
games647, line 379: Could you check if this one gets shaded into the plugin? It might should be provided if we only reference on it.

games647 (changes requested):
Looks good so far

Smart123s:
> I rather was talking about the concept of having change based migrations.

Then I misunderstood you, I thought you were talking about the library.

> I don't think we need a library for that.

Me neither.

> This might be a bit overkill

It'd make code cleaner, and it'd also make potential future migrations easier, so I think it's fine.

> I guess that's the best option to migrate the old data while using the new structure for new players. We could later drop the backwards compatibility by migrating all nullable values to false.

This will be the way then.

core/src/main/java/com/github/games647/fastlogin/core/storage/SQLStorage.java
```diff
@@ -90,6 +97,57 @@ public void createTables() throws SQLException {
...
+                + "ADD COLUMN Floodgate BOOLEAN NOT NULL DEFAULT 0";
+        final String dbRevStatement = "ALTER TABLE `" + PREMIUM_TABLE + "` "
+                + "ADD COLUMN DBVersion INTEGER NOT NULL DEFAULT 1";
+
+        try (Connection con = dataSource.getConnection()) {
+            // todo: add the two columns with a single connection to the database
+            try (PreparedStatement saveStmt = con.prepareStatement(floodgateStatement)) {
+                saveStmt.execute();
+            }
+            try (PreparedStatement saveStmt = con.prepareStatement(dbRevStatement)) {
```
games647, line 127: You don't need a prepared statement here, because there are no placeholders.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/ProtocolLibListener.java
```diff
@@ -254,4 +267,65 @@ private String getUsername(PacketContainer packet) {
...
+        if (floodgatePlayer == null) {
+            return true;
+        }
+
+        // kick the player, if necessary
+        Channel channel = handler.getChannel(packetEvent.getPlayer());
+        AttributeKey<String> kickMessageAttribute = AttributeKey.valueOf("floodgate-kick-message");
+        String kickMessage = channel.attr(kickMessageAttribute).get();
+        if (kickMessage != null) {
+            packetEvent.getPlayer().kickPlayer(kickMessage);
```
games647, line 312: Use the player variable from above directly.

games647 (changes requested):

Smart123s:
I've moved the ProtocolLib fix to https://github.com/games647/FastLogin/pull/876 so I won't have to rebase ProtocolLib related commits when working on the database. As long as the fix gets merged before the database modifications, things should be fine.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/ProtocolLibListener.java
Smart123s: Moved to & fixed in #876

bukkit/pom.xml
Smart123s: Moved to & fixed in #876

Smart123s:
The migrator works, but the code is a mess in my opinion. 

I think that it'd be better to refactor the code, and make SQLStorage a kind of connection manager (initialize connection to the db) and make a separate class for every table. There classes would access `dataSource` from the common SQLStorage class. Should I go for it?

<hr>

Additionally, is the `ReentrantLock` in SQLIteStorage needed? I think that Hikari/JDBC makes sure that only one connection is open to the database on its own. By accident I forgot to close a `Connection` in MigrationManager after executing a query, and and upon the first login attempt (or the first call to  `dataSource.getConnection();`) I got the following error:

```
[16:59:59] [Craft Scheduler Thread - 6 - FastLogin/ERROR]: [FastLogin] Failed to query profile: .Smart123ss
java.sql.SQLTransientConnectionException: FastLogin - Connection is not available, request timed out after 30010ms.
	at fastlogin.hikari.pool.HikariPool.createTimeoutException(HikariPool.java:696) ~[FastLoginBukkit.jar:?]
	at fastlogin.hikari.pool.HikariPool.getConnection(HikariPool.java:197) ~[FastLoginBukkit.jar:?]
	at fastlogin.hikari.pool.HikariPool.getConnection(HikariPool.java:162) ~[FastLoginBukkit.jar:?]
	at fastlogin.hikari.HikariDataSource.getConnection(HikariDataSource.java:100) ~[FastLoginBukkit.jar:?]
	at com.github.games647.fastlogin.core.storage.SQLStorage.loadProfile(SQLStorage.java:107) ~[FastLoginBukkit.jar:?]
	at com.github.games647.fastlogin.core.storage.SQLiteStorage.loadProfile(SQLiteStorage.java:83) ~[FastLoginBukkit.jar:?]
	at com.github.games647.fastlogin.core.shared.FloodgateManagement.run(FloodgateManagement.java:83) ~[FastLoginBukkit.jar:?]
	at org.bukkit.craftbukkit.v1_19_R1.scheduler.CraftTask.run(CraftTask.java:101) ~[paper-1.19.1.jar:git-Paper-110]
	at org.bukkit.craftbukkit.v1_19_R1.scheduler.CraftAsyncTask.run(CraftAsyncTask.java:57) ~[paper-1.19.1.jar:git-Paper-110]
	at com.destroystokyo.paper.ServerSchedulerReportingWrapper.run(ServerSchedulerReportingWrapper.java:22) ~[paper-1.19.1.jar:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]
[16:59:59] [Craft Scheduler Thread - 6 - FastLogin/WARN]: [FastLogin] Plugin FastLogin v1.12-SNAPSHOT-a47b879 generated an exception while executing task 17
java.lang.NullPointerException: Cannot invoke "com.github.games647.fastlogin.core.StoredProfile.isSaved()" because "this.profile" is null
	at com.github.games647.fastlogin.core.shared.FloodgateManagement.run(FloodgateManagement.java:84) ~[FastLoginBukkit.jar:?]
	at org.bukkit.craftbukkit.v1_19_R1.scheduler.CraftTask.run(CraftTask.java:101) ~[paper-1.19.1.jar:git-Paper-110]
	at org.bukkit.craftbukkit.v1_19_R1.scheduler.CraftAsyncTask.run(CraftAsyncTask.java:57) ~[paper-1.19.1.jar:git-Paper-110]
	at com.destroystokyo.paper.ServerSchedulerReportingWrapper.run(ServerSchedulerReportingWrapper.java:22) ~[paper-1.19.1.jar:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]
```

*Note: I've rebased the code multiple times since I encountered this issue (and fixed it), so I can't provide the exact lines and context where the issue occurred. I didn't test MySQL for this issue.*

core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java
```diff
@@ -138,8 +152,14 @@ private boolean checkPremiumName(S source, String username, StoredProfile profil
     private boolean checkNameChange(S source, String username, Profile profile) {
         //user not exists in the db
         if (core.getConfig().get("nameChangeCheck", false)) {
-            StoredProfile storedProfile = core.getStorage().loadProfile(profile.getId());
+            StoredProfile storedProfile = core.getAuthStorage().loadProfile(profile.getId());
             if (storedProfile != null) {
+                if (storedProfile.isFloodgate()) {
+                    core.getPlugin().getLog()
+                            .warn("Player {} is already stored by FastLogin as a Bedrock Edition player.", username);
```
games647, line 159:
```suggestion
                            .info("Player {} is already stored by FastLogin as a Bedrock Edition player.", username);
```

Sounds more informative than warning the server owner.

core/src/main/java/com/github/games647/fastlogin/core/shared/JoinManagement.java
```diff
@@ -50,19 +50,33 @@ public JoinManagement(FastLoginCore<P, C, ?> core, AuthPlugin<P> authHook, Bedro
...
+        if (profile == null) {
+            return;
+        }
+
+        if (!profile.isFloodgateMigrated()) {
+            profile.setFloodgate(false);
+            core.getPlugin().getLog().info(
+                    "Player {} will be migrated to the v2 database schema as a JAVA user", username);
+        } else if (profile.isFloodgate()) {
+            core.getPlugin().getLog().warn("Player {} is already stored by FastLogin as a Bedrock Edition player",
```
games647, line 75:
```suggestion
            core.getPlugin().getLog().info("Player {} is already stored by FastLogin as a Bedrock Edition player",
```

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
```diff
@@ -76,11 +76,38 @@ public void run() {
...
             return;
         }
 
-        profile = core.getStorage().loadProfile(username);
+        profile = core.getAuthStorage().loadProfile(username);
+        if (!profile.isSaved()) {
+            // linked players are stored as Java (= not Floodgate) players
+            profile.setFloodgate(!isLinked);
+        } else {
+            if (!profile.isFloodgateMigrated()) {
```
games647, line 88: Please put the non-inverted case first. It's easier to follow the flow.

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
```diff
@@ -76,11 +76,38 @@ public void run() {
...
+                    core.getPlugin().getLog().info(
+                            "Player {} will be migrated to the v2 database schema as a JAVA (linked Floodgate) user",
+                            username);
+                } else {
+                    profile.setFloodgate(true);
+                    core.getPlugin().getLog().info(
+                            "Player {} will be migrated to the v2 database schema as a Floodgate user", username);
+                }
+            } else if (!profile.isFloodgate() && !isLinked) {
+                core.getPlugin().getLog().warn("Player {} is already stored by FastLogin as a Java Edition player",
```
games647, line 100:
```suggestion
                core.getPlugin().getLog().info("Player {} is already stored by FastLogin as a Java Edition player",
```

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
```diff
@@ -76,11 +76,38 @@ public void run() {
...
+                        username);
+                return;
+            } else if (profile.isFloodgate() && isLinked) {
+                core.getPlugin().getLog()
+                        .warn("Player {} is already stored by FastLogin as a non-linked Bedrock Edition player",
```
games647, line 105:
```suggestion
                        .info("Player {} is already stored by FastLogin as a non-linked Bedrock Edition player",
```

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
```diff
@@ -76,11 +76,38 @@ public void run() {
...
-        if (core.getStorage() == null) {
+        if (core.getAuthStorage() == null) {
             return;
         }
 
-        profile = core.getStorage().loadProfile(username);
+        profile = core.getAuthStorage().loadProfile(username);
+        if (!profile.isSaved()) {
+            // linked players are stored as Java (= not Floodgate) players
+            profile.setFloodgate(!isLinked);
```
games647, line 86: Why do you considered this approach? I feel like setting this to `false` could lead to the same issues. Is there any issue leaving it true and then read the configuration instead? This would make it easier to change configuration files.

Maybe we could make this field also a tristate.

core/src/main/java/com/github/games647/fastlogin/core/storage/MigratableStorage.java
```diff
@@ -0,0 +1,43 @@
...
+package com.github.games647.fastlogin.core.storage;
+
+public interface MigratableStorage {
+
+    /**
+     * Get an SQL statement to migrate the table to the next version.
+     * @param currentVersion The current version of the table
+     * @return an SQL statement to migrate the table to the next version
+     */
+    String getMigrationStatement(int currentVersion);
```
games647, line 35: Wouldn't it be better to have the migrations as files or separate Java classes? For example for the case a migration performs multiple statements.

core/src/main/java/com/github/games647/fastlogin/core/storage/MigrationManager.java
```diff
@@ -0,0 +1,124 @@
...
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+public class MigrationManager {
+
+    protected static final String MIGRATION_TABLE = "migrations";
+    protected static final String CREATE_TABLE_STMT = "CREATE TABLE IF NOT EXISTS `" + MIGRATION_TABLE + "` ("
```
games647, line 39: Migration approaches often use one row for each migration that you can individually rollback. I think this could be overkill for this project, but at least we should short comment what this approach does or explain the table scheme.

core/src/main/java/com/github/games647/fastlogin/core/storage/MigrationManager.java
```diff
@@ -0,0 +1,124 @@
...
+
+    protected void migrateTable(MigratableStorage table) {
+        int initialVersion = getTableVersion(table.getTableName());
+
+        // special case: table premium was created before the migration manager
+        if (initialVersion == 0 && "premium".equals(table.getTableName()) && table.tableExists()) {
+            initialVersion = 1;
+        }
+
+        for (int i = initialVersion; i < table.getLatestTableVersion(); i++) {
```
games647, line 101: We should better differentiate between what `latest` or `getTableVersion` means. Maybe something like `getCurrentVersion` and `getRequiredVersion`. Or add a comment that makes it clear.

games647 (changes requested):
Looks very good so far.

core/src/main/java/com/github/games647/fastlogin/core/storage/AuthStorage.java
```diff
@@ -35,6 +35,4 @@ public interface AuthStorage {
     StoredProfile loadProfile(UUID uuid);
 
     void save(StoredProfile playerProfile);
-
-    void close();
```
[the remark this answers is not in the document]
Smart123s: I moved it from AuthStorage to

https://github.com/games647/FastLogin/blob/06418b637ce27d31e1d4365a5d56d491add3142b/core/src/main/java/com/github/games647/fastlogin/core/storage/SQLStorage.java#L74-L76

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
Smart123s: Linked Floodgate players are 99% Java players: they have the same name (without prefix), the same UUID, inventory, etc. The only difference is that they are present in the Floodgate API and they can't reply to premium java login requests.
I decided to keep it `false`, because they are the same player as their Java part, and I wanted to avoid storing one player as two entries in the database.

core/src/main/java/com/github/games647/fastlogin/core/storage/MigrationManager.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        return -1;
+    }
+
+    protected void migrateTable(MigratableStorage table) {
+        int initialVersion = getTableVersion(table.getTableName());
+
+        // special case: table premium was created before the migration manager
+        if (initialVersion == 0 && "premium".equals(table.getTableName()) && table.tableExists()) {
```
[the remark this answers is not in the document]
Smart123s: Resolved in 0dae497b9d594271acdf62a0fa767beb112869c9

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
Smart123s: Resolved in 35c289d26d115ffaaafdc02dc09f36f1b02ce175

core/src/main/java/com/github/games647/fastlogin/core/storage/MigrationManager.java
Smart123s: Changed in e401103bc93a0d365499887672a99b46a3bcb752

core/src/main/java/com/github/games647/fastlogin/core/storage/MigrationManager.java
Smart123s: Is 69df3a58d79c054c4f38c1bf4cb9692ca2986624 enough?

core/src/main/java/com/github/games647/fastlogin/core/storage/AuthStorage.java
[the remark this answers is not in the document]
games647: Well I think it's part of a functionality of the interface and not of the specific implementation of SQLStorage. To be precise, cases like FileStorage then should also have a close method.

core/src/main/java/com/github/games647/fastlogin/core/storage/MigratableStorage.java
Smart123s: I have added https://github.com/games647/FastLogin/pull/709/commits/b59cc66d1091c3836d8119636260d7c30573250c, could you please check it?

core/src/main/java/com/github/games647/fastlogin/core/storage/AuthStorage.java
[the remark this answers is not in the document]
Smart123s: Is 3ffb9c7b18c8cb730078c75943cff407a615888b ok, or should I remove the close method from SQLStorage and reference `storage.dataSource` directly in SQLAuthStorage?

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
Smart123s: I've added 5d994d8, but it's still a mess.

Let's say we have a player called Steve, who plays from Java, and linked Floodgate too. What should the value of Floodgate be in our database? Linked or Java? Both of them are correct. Names are unique in the database, so there's no way to have two row for one player.

games647: I think I'm lost. Why would we need two rows? You mean for the case Floodgate and Java profiles are associated to different accounts/persons? This would make it very difficult, because the auth plugin has only one single account often based on the player name. However, auth plugins with UUID support would in fact be different. 

The reason I thought name is unique, is because the only a single user can have this Mojang account name. If a user joins with the same name, but a different UUID, this would mean that the old data name is now invalid.

Smart123s: I want to keep names unique, changing it would cause a lot of headaches.

What I tried to explain is that Java and their linked Floodgate accounts have the same name (and UUID), and I don't know how to save them in the database.

For example, I join a brand new server from my PC with my Java account. Then I get saved in the database as `Smart123s`, `FloodgateState.JAVA`. Then I get on the bus, grab my phone, and join the server with my linked Floodgate account. My name is still `Smart123s`, because I'm linked, but I'm no longer `FloodgateState.JAVA`, because I'm joining from bedrock this time. Should the database be changed to `Smart123s`, `FloodgateState.LINKED`? And then what happens if I join from Java again?

That's what I initially wanted to avoid by storing linked Floodgate players as Java players.

games647: >Should the database be changed to Smart123s, FloodgateState.LINKED? And then what happens if I join from Java again?

Well we could keep it as linked as it would mean (Floodgate+Java), so it doesn't represent the current session, but the account/person that is behind it.

Smart123s: Ok, but there are two minor caveats. Firstly, it's impossible to check if a new Java player has a linked Floodgate account until he joins with it. And it's also impossible to detect if a Floodgate account was unlinked. Technically, we could send API requests to Geyser to check for these, but it'd slow down the login process without any actual benefit.

Anyways, it's gonna be a busy week/weekend for me, but I'll check on the code to make sure it works as we discussed above.

Smart123s:
I have tested the migration from 7425df8 to 1d0e089 on paper 1.19.2 with SQLite and MariaDB and encountered no issues with it.

Smart123s:
@games647 The code is ready for review (and merge?), when you have some spare time. No need to rush.

Smart123s:
Hi, I know it's been a long time, but I kind of want to rewrite some parts.

While I remember that we had discussions about the Migrations table, and wether or not it's neccesarry. I've looked at the code of AuthMe, and they don't use a Migrations table. Here's what they use:

https://github.com/AuthMe/AuthMeReloaded/blob/master/src/main/java/fr/xephi/authme/datasource/MySQL.java#L191-L301
https://github.com/AuthMe/AuthMeReloaded/blob/master/src/main/java/fr/xephi/authme/datasource/MySqlMigrater.java

Do you think something similar would be sufficient here?

The commit history could also take some squashing here and there. (and I'll also sign the commits you've pushed).

I saw that there's an [integration-test-containers](https://github.com/games647/FastLogin/tree/integration-test-containers) branch. Either that should be completed, or at least some JUnit tests should be added to make sure that database migrations won't break anytime soon.

games647:
>The commit history could also take some squashing here and there. (and I'll also sign the commits you've pushed).

**Warning I lost some commits during rebase. Maybe you could restore your version or I'll try to dig into my reflog.** 

>Do you think something similar would be sufficient here?

To be honest: This migration concept could be a little too much for a single database change (atm). The idea of migrations comes from my experience with web projects where I saw it quite often with big benefits for multiple contributors (separating logical changes to different files) and being able to rollback. In the end, I'm fine with the AuthMe system too.

>I saw that there's an [integration-test-containers](https://github.com/games647/FastLogin/tree/integration-test-containers) branch.

Yeah the idea was to test core functionality against a full Minecraft server. However, this is very hard to setup and optimize. Nevertheless, it could be very helpful for verifying aspects like login support. 

Regarding the database system, this could be a lot easier and a good idea. By separating the encapsulating the database implementation from the Bukkit API we could be fine with only running a MySQL/MariaDB image.

Smart123s:
Let's not forget to check this on Velocity since https://github.com/games647/FastLogin/pull/1071 got merged. This PR only changes code in core, so it's possible, that it'll work without any modifications, but who knows? I'll be pretty busy nowdays, I'll report back once I have some time.

core/src/main/java/com/github/games647/fastlogin/core/shared/FloodgateManagement.java
```diff
@@ -119,15 +153,12 @@ public void run() {
             }
         }
 
-        if (!isRegistered && !isAutoAuthAllowed(autoRegisterFloodgate)) {
+        if ((!isRegistered && !isAutoAuthAllowed(autoRegisterFloodgate))
+                || (profile.getFloodgate() == FloodgateState.LINKED && !isLinked)
```
games647, line 157: Could you please explain the different cases and why they are made. For me it's hard to understand what is actually happening here. Maybe you can extract them to readable methods if that helps.

games647 (changes requested):
Finally I have some free time so I looked over the code. I would like to merge it now and revive some work on the project. I only inverted some ifs, because for me personally reading the non-inverted-case is easier to read than having shorter code.

In the end, I only have trouble to understand the if-cases in FloodgateManagement that would like to have resolved before merging it. Thank you in advance for the help.

Smart123s:
> Finally I have some free time so I looked over the code. I would like to merge it now and revive some work on the project. I only inverted some ifs, because for me personally reading the non-inverted-case is easier to read than having shorter code.
> 
> In the end, I only have trouble to understand the if-cases in FloodgateManagement that would like to have resolved before merging it. Thank you in advance for the help.

Those if statements got really complicated, that's for sure. Some months ago I tried looking into it to simplify the code, but I couldn't really do anything about it.

I have force pushed those changes as a fixup of Differentiate Floodgate players during login & rebased the branch on main, so you may have to re-sign your commits.

It's nice to hear from you after all this time! Thanks for remembering, and looking into my code after so long Hope you are alright!

games647 (approved):

Issue #736, opened 2022-02-25
Floodgate players won't get autologed


paj1s:
### What happened?

Bedrock player joins server and is not autologed nor autoregistered.

### What did you expect?

Bedrock player joins server and will be autoregistered and autologed according to the config.

### Steps to reproduce

I have waterfall with geyser,floodgate and fastlogin bungee connected to mysql
and paper lobby server with authme, floodgate and fastlogin also connected to mysql
everything is working fine except for the bedrock.
Im not using floodgate prefix.

### Plugin list

floodgate,authme,fastlogin,floodgate,protocollib,luckperms,vault...

### Configuration file

https://www.toptal.com/developers/hastebin/umajegilon.yaml

### Server log

```shell
lobby log
[15:38:04 INFO]: UUID of player pajis4452 is 00000000-0000-0000-0009-01fb38b30c13
[15:38:04 INFO]: pajis4452[/78.80.120.71:48265] logged in with entity id 896 at ([hub]8.724118516533247, 81.0, 8.49924104612512)
[15:38:04 INFO]: pajis4452 se připojil
[15:38:05 INFO]: [FastLogin] No on-going login session for player: CraftPlayer{name=pajis4452} with ID 78.80.120.71:48265
[15:38:09 INFO]: UUID of player pajis4452 is 00000000-0000-0000-0009-01fb38b30c13
[15:38:09 INFO]: pajis4452 lost connection: Disconnected
[15:38:09 INFO]: pajis4452 se odpojil
waterfall log
[15:38:02 INFO] [Geyser-BungeeCord]: /78.80.120.71:38020 tried to connect!
[15:38:03 INFO] [Geyser-BungeeCord]: Player connected with username pajis4452
[15:38:04 INFO] [Geyser-BungeeCord]: pajis4452 (logged in as: pajis4452) has connected to the Java server
[15:38:04 INFO]: [/78.80.120.71:0] <-> InitialHandler has connected
[15:38:04 INFO] [FastLogin]: Incoming login request for pajis4452 from /78.80.120.71:0
[15:38:04 INFO] [FastLogin]: Handling player pajis4452
[15:38:04 INFO] [floodgate]: Floodgate player logged in as pajis4452 joined (UUID: 00000000-0000-0000-0009-01fb38b30c13)
[15:38:04 INFO]: [pajis4452|/78.80.120.71:0] <-> ServerConnector [Lobby] has connected
[...]

PR #737, opened 2022-02-25
Fix 'no-conflict' option of 'autoLoginFloodgate' in config

Smart123s:
### Summary of your change
Added checks for missing `no-conflict` checks to this code for both Bukkit & Bungee

https://github.com/games647/FastLogin/blob/4c514c269b46bb6d28b9e34e6eecb3448eefdedd/bungee/src/main/java/com/github/games647/fastlogin/bungee/task/FloodgateAuthTask.java#L58-L60

Instead of simply adding the missing two lines, I've done a bit of a code clean up. I've merged the checks of the code mentioned above and `isAutoRegisterAllowed()` into a common function. I've made 2 commits, to make it easier to understand what I did and why.

### Related issue
Fixes https://github.com/games647/FastLogin/issues/736


games647 (approved):
!LGTM

Thank you

PR #750, opened 2022-03-13
Fix License header

Smart123s:
### Summary of your change
Updated license year
Re-run `mvn license:format`
Added `mvn license:check` to the maven build goals, so the build will fail if license headers are incorrect.

### Related issue
none


games647 (approved):
Good catch. Thank you.

PR #779, opened 2022-04-07
Add a code formatter

Smart123s:
### Summary of your change

Added a maven plugin that can auto format the code & check if the code is formatted properly.
I've tried to create a code style that's similar to the current style of FastLogin, although it wasn't easy, because it doesn't follow a consistent code style. For example, in some files, lines are broken at char 80, where in other, they are broken at char 120. See the commit message of https://github.com/games647/FastLogin/commit/3fe90a6b0e14021abf11e990aae8ec5959379bb5 for details on the proposed formatter config. If you have any idea on how to improve it, please let me know.

### Related issue
I just hate it that i don't know how to format the code.


games647:
>If you don't want to use a formatter, just close this PR, it's just an idea. If you don't like my formatter, replace it. I had to propose something, I couldn't leave it empty.

No it's a good practice and we should add something like this. 

Does it only lint or does it also automatically format the code? How good is the compatibility with other IDEs like IntelliJ and Netbeans. Checkstyle is also widely known, but I don't know the current state of the tools.

Smart123s:
I've tried checkstyle, imported the `sun_checks.xml` to Intelij and eclipse, ran code formatter in both of them. The I ran `mvn checkstyle:check`, and got

```
There are 883 errors reported by Checkstyle 8.29 with sun_checks.xml ruleset.
```

Regardless, this seems to be a really good tool, I think this one should be used, but it'll take some time, so I just wanted to give you an update.

PR #811, opened 2022-06-09
Support for transparent proxies / DNAT

### LINKED ISSUE #810: Tunnelled server connection -> Invalid Session
### What happened?

Basically, I tried installing FastLogin, ProtocolLib and LoginSecurity on my server. The server has run in online mode before, so I needed FastLogin in order to switch to offline mode (what I mean by that that I enabled premiumUuids so the UUIDs stay the same for the original players).

I run the server on a machine in my local network. But I can't easily just forward ports to this machine. So I used my other (public-facing) VPS to act as a routing reverse proxy for my server.

Basically, my setup looks like this.

Player ---> Proxy server <==(10.25.25.1)==[OpenVPN Tunnel]==(10.25.25.2)==> Minecraft Server

The proxy server does not run any minecraft-related software. All it does is run OpenVPN server (to which the MC server connects to). Additionally, to achieve the port forwarding, I used the following iptables rules:

```
iptables -t nat -A PREROUTING -p tcp --dport 25565 -j DNAT --to-destination 10.25.25.2:25565
iptables -t nat -A POSTROUTING -p tcp --dport 25565 -j MASQUERADE
```

To the actually relevant part. While the server runs in online mode, everything works flawlessly. When I installed FastLogin and LoginSecurity, none of my friends could log in. They all received the "Invalid Session" message. What's more interesting is that I was able to log in just fine using the same tunnelled connection.

On a side note, I run the server in pufferpanel running thorough docker. AFAIK docker also does it's own routing so that might also be an issue.

I also tried debugging the plugin through IntelliJ IDEA, but there everything worked just fine. I even tried to have it use the same VPN tunnel, and it worked as expected.

Honestly, I don't know what to do, I was just hoping you might have more insight in the role routing plays in the protocol (i.e. if the SNAT has some negative impact or something like that).

### What did you expect?

I expected the tunnel to work fine, just like it used in online mode.

### Steps to reproduce

On machine A (facing internet):
1. Setup OpenVPN server with `ifconfig 10.25.25.1 10.25.25.2` directive
2. Setup the iptables rules:
    ```
    iptables -t nat -A PREROUTING -p tcp --dport 25565 -j DNAT --to-destination 10.25.25.2:25565
    iptables -t nat -A POSTROUTING -p tcp --dport 25565 -j MASQUERADE
    ```

On machine B:
1. Setup OpenVPN client to Machine A with `ifconfig 10.25.25.2 10.25.25.1`
3. Install docker and docker-compose
4. Create `docker-compose.yml`
    ```
    version: '3.8'

    services:
      pufferpanel:
        container_name: puffer
        image: pufferpanel/pufferpanel:java
        volumes:
          - ./puffer-data:/var/lib/pufferpanel
          - ./puffer-config:/etc/pufferpanel
        ports:
          - "8080:8080"
          - "25565:25565"
    ```
5. Start the container (`docker-compose up -d`)
6. Install the minecraft server using pufferpanel (use the provided template for spigot)
7. Install all the plugins mentioned below
8. Use the provided config for FastLogin
9. Use the following config for LoginSecurity (minimized):
    ```
    register:
      required: true
      captcha: false
      confirm-password: false
    login:
      max-tries: 5
      username-match-exact: true
    password:
      min-length: 6
    password-max-length: 32
    join:
      blindness: true
      location: SPAWN
      hide-inventory: true
      hide-inventory-safe: false
    username:
      filter-special-chars: true
      min-length: 3
      max-length: 16
    command-shortcut:
      enabled: false
      login: /l
      register: /reg
    updater:
      enabled: false
      channel: BETA
    session-timeout: -1
    login-timeout: 120
    login-message-delay: 10
    language: en_us
    ```
11. Change the `settings/timeout-time` in spigot.yml to 80 (With the default 60 spigot watchdog thread throws errors on stop)
12. Start the server through pufferpanel

NOTE: This is just a rough guide on how I set up the server, but it does not guarantee the bug can be reproduced, as even I wasn't able to reproduce it reliably.

### Plugin list

ProtocolLib, CrisTreeCapitator, FastLogin, LoginSecurity, ServerBackup

### Configuration file

https://www.toptal.com/developers/hastebin/figacelabi.yaml

### Server log

```shell
https://www.toptal.com/developers/hastebin/tisizewaba.yaml
```

### Plugin version

1075

### Platform

Spigot

### Relevance

- [X] I tried the latest build (build refers to development builds not necessary a release version)
- [X] I checked for existing tickets -
If there are, please vote them with a thumps reaction and not create new ones

Enginecrafter77:

### Proposed changes
This pull request addresses #810 and other possibly relates cases of servers being hidden behind transparent reverse proxy.

The basic motivation behind this change is this scenario. Imagine you had a minecraft server. You cannot forward ports on your firewall (e.g. a bad ISP). So you opt for an external server which will act as a reverse proxy for your server. The users of your server would connect to the proxy server, which will transparently route the data to your server (e.g. using a VPN or SSH tunnel). 

The problem is that in some cases, FastLogin will not work due to the way it invokes the `hasJoined` method on mojang session servers. The default CraftAPI implementation, along with other data, includes the `ip` parameter in the request. This, however, is a problem since mojang has stored the true IP of the client, while the server sends the IP of the proxy server, and thus the authentication fails with "Invalid Session".

The way this pull request resolves the issue is by leaving out the `ip` parameter altogether. According to [wiki.vg page about the protocol authentication](https://wiki.vg/Protocol_Encryption#Authentication), the `ip` parameter is optional. Leaving out the parameter skips the check which was causing the "Invalid Session", and thus allows players to play on your server using transparent TCP proxies.

Regarding the security concerns of this implementation, the user opting for such measures should already be aware that by using a transparent proxy, all the user IP addresses will be the same, and therefore things like IP bans and/or plugins that rely on unique IP per player might not work as expected. Additionally, since the user is already aware of the fact that the player's IP as seen by the server is not the real IP of the player, they do not need mojang to check the player IP if they already know it is not the real one.

The feature implemented by this pull request is deactivated by default and can be enabled through the FastLogin config file. Older versions of the config file missing the entry will NOT cause problems since the key defaults to false when not present. That means this feature will NOT impact existing users of FastLogin, but will merely enable interested users to turn it on.

### Related issue
Fixes #810 and related cases

### DIFF
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
@@ -141,10 +141,7 @@ public class VerifyResponseTask implements Runnable {
         try {
             MojangResolver resolver = plugin.getCore().getResolver();
             InetAddress address = socketAddress.getAddress();
-
-            // TODO edit marker
-            Optional<Verification> response = VerifyResponseTask.hasJoined(resolver, requestedUsername, serverId, address);
-
+            Optional<Verification> response = resolver.hasJoined(requestedUsername, serverId, address);
             if (response.isPresent()) {
                 Verification verification = response.get();
                 plugin.getLog().info("Profile {} has a verified premium account", requestedUsername);
@@ -174,83 +171,6 @@ public class VerifyResponseTask implements Runnable {
         } catch (IOException ioEx) {
             disconnect("error-kick", "Failed to connect to session server", ioEx);
         }
-    }
-
-    /**
-     * A reimplementation of {@link MojangResolver#hasJoined(String, String, InetAddress)} using various crude reflection-based hacks
-     * to access the protected code. The significant difference is that unlike in the CraftAPI implementation, which sends the "ip" parameter
-     * when the hostIp parameter is an IPv4 address, but skips it for IPv6, this implementation ommits the "ip" parameter also for IPv4, effectively
-     * enabling transparent proxies to work.
-     *
-     * TODO Reimplement as MojangResolver subclass overriding hasJoined method -> "ProxyAgnosticMojangResolver" perhaps?
-     *
-     * @param resolver The mojang resolver object the method will operate on
-     * @param username The literal username of the player
-     * @param serverHash The computed server hash sent to mojang session servers
-     * @param hostIp The host IP address - not used, kept to maintain similar signature
-     * @return An optional object containing the verification information, if any. If there is no verification information, the session servers consider the join invalid.
-     * @throws IOException When an error occurs during the HTTP communication
-     * @author games647, Enginecrafter77
-     */
-    public static Optional<Verification> hasJoined(MojangResolver resolver, String username, String serverHash, InetAddress hostIp) throws IOException {
-        /*String url;
-        if (hostIp instanceof Inet6Address) {
-            url = String.format("https://sessionserver.mojang.com/session/minecraft/hasJoined?username=%s&serverId=%s", username, serverHash);
-        } else {
-            String encodedIP = URLEncoder.encode(hostIp.getHostAddress(), StandardCharsets.UTF_8.name());
-            url = String.format("https://sessionserver.mojang.com/session/minecraft/hasJoined?username=%s&serverId=%s&ip=%s", username, serverHash, encodedIP);
-        }*/
-        String url = String.format("https://sessionserver.mojang.com/session/minecraft/hasJoined?username=%s&serverId=%s", username, serverHash);
-
-        try
-        {
-            Class<?> inputStreamActionClass = Class.forName("com.github.games647.craftapi.resolver.AbstractResolver$InputStreamAction");
-            Method getConnectionMethod = AbstractResolver.class.getDeclaredMethod("getConnection", String.class);
-            Method parseRequestMethod = AbstractResolver.class.getDeclaredMethod("parseRequest", HttpURLConnection.class, inputStreamActionClass);
-            Method readJsonMethod = AbstractResolver.class.getDeclaredMethod("readJson", InputStream.class, Class.class);
-
-            getConnectionMethod.setAccessible(true);
-            parseRequestMethod.setAccessible(true);
-            readJsonMethod.setAccessible(true);
-
-            HttpURLConnection conn = (HttpURLConnection)getConnectionMethod.invoke(resolver, url);
-            int responseCode = conn.getResponseCode();
-
-            Verification result = null;
-            if(responseCode != 204)
-            {
-                AbstractResolverAdapter.InputStreamActionAdapter<Verification> action = new AbstractResolverAdapter.InputStreamActionAdapter<Verification>() {
-                    @Override
-                    public Verification useStream(InputStream inp) throws IOException
-                    {
-                        try
-                        {
-                            return (Verification)readJsonMethod.invoke(resolver, new Object[] {inp, Verification.class});
-                        }
-                        catch(ReflectiveOperationException exc)
-                        {
-                            throw new IOException("Reflective method access failed", exc);
-                        }
-                    }
-                };
-                result = (Verification)parseRequestMethod.invoke(resolver, new Object[] {conn, action});
-            }
-            return Optional.ofNullable(result);
-        }
-        catch(ReflectiveOperationException exc)
-        {
-            throw new RuntimeException("Error occured in reflective hacks to MojangResolver", exc);
-        }
-    }
-
-    /**
-     * An adapter class used to make the InputStreamAction interface accessible to us.
-     * I know, it's a crude and disgusting solution, but bear with me, it's just temporary.
-     */
-    public static class AbstractResolverAdapter extends AbstractResolver
-    {
-        @FunctionalInterface
-        protected static interface InputStreamActionAdapter<R> extends AbstractResolver.InputStreamAction<R> {}
     }
 
     private void setPremiumUUID(UUID premiumUUID) {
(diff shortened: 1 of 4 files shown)


bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
```diff
@@ -162,8 +165,8 @@ private void verifyResponse(BukkitLoginSession session) {
             } else {
                 //user tried to fake an authentication
                 disconnect("invalid-session",
-                    "GameProfile {0} ({1}) tried to log in with an invalid session ServerId: {2}",
-                    session.getRequestUsername(), socketAddress, serverId);
+                    String.format("GameProfile %s (%s) tried to log in with an invalid session. ServerId: %s",
```
games647, line 168: SLF4J uses {} for logging, instead of {numeric} which was missing in the migration to it. Maybe it's better to use the formatting of the logger than to do it manually.

core/src/main/java/com/github/games647/fastlogin/core/mojang/ProxyAgnosticMojangResolver.java
```diff
@@ -0,0 +1,64 @@
...
+	@Override
+	public Optional<Verification> hasJoined(String username, String serverHash, InetAddress hostIp) throws IOException
+	{
+		String url = String.format("https://sessionserver.mojang.com/session/minecraft/hasJoined?username=%s&serverId=%s", username, serverHash);
+
+		HttpURLConnection conn = this.getConnection(url);
+		int responseCode = conn.getResponseCode();
+
+		Verification verification = null;
+		if(responseCode != 204)
```
games647, line 54: Please use the HTTPURLConnection constant. 204 is just a magic number, which seems to better behind a name.

core/src/main/resources/config.yml
```diff
@@ -152,6 +152,17 @@ forwardSkin: true
 # If they still want to invoke the command, they have to invoke /premium again
 premium-warning: true
 
+# When set to true, enables the use of alternative session resolver which does not send the server IP
+# to mojang session servers. This setting might be useful when you are trying to run the server via a
+# transparent reverse proxy or some other form of DNAT. As far as security goes, this setting has
+# negligible to no security impact.
```
games647, line 158: Might be necessary to mention this is only related to Spigot protocollib feature and corresponds to the `prevent-proxy=false` in `server.properties`.

From Mojang this feature was likely introduced to prevent any form of authentication proxies. So client modifications that allow to use Minecraft authentication provided by a third-party without actually receiving the credentials or proxying all traffic.

games647 (changes requested):

core/src/main/java/com/github/games647/fastlogin/core/mojang/ProxyAgnosticMojangResolver.java
```diff
@@ -0,0 +1,64 @@
...
+ * The significant difference is that unlike MojangResolver from the CraftAPI implementation, which sends the "ip" parameter
+ * when the hostIp parameter is an IPv4 address, but skips it for IPv6, this implementation leaves out the "ip" parameter
+ * also for IPv4, effectively enabling transparent proxies to work.
+ * @author games647, Enginecrafter77
+ */
+public class ProxyAgnosticMojangResolver extends MojangResolver {
+	@Override
+	public Optional<Verification> hasJoined(String username, String serverHash, InetAddress hostIp) throws IOException
+	{
+		String url = String.format("https://sessionserver.mojang.com/session/minecraft/hasJoined?username=%s&serverId=%s", username, serverHash);
```
games647, line 48: Please make this string a constant.

games647:
Looks good so far.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
Enginecrafter77: Yeah this is just a left-over from my debugging since SLF4J would not for some reason expand the parameters and I wanted to see them so I just put this in place. I will put it back as it was.

core/src/main/java/com/github/games647/fastlogin/core/mojang/ProxyAgnosticMojangResolver.java
Enginecrafter77: Noted and will fix. In my defense, I just used whatever was in the original CraftAPI code, I wasn't aware of the constants.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/VerifyResponseTask.java
Enginecrafter77: Resolved in https://github.com/games647/FastLogin/pull/811/commits/6413ca4d109d942d2c461916263413218b0551b4

core/src/main/java/com/github/games647/fastlogin/core/mojang/ProxyAgnosticMojangResolver.java
Enginecrafter77: Resolved in https://github.com/games647/FastLogin/pull/811/commits/edd82f7978fdb9097beaa1424ee19df5342816e9

core/src/main/resources/config.yml
Enginecrafter77: I have attempted to rephrase it in https://github.com/games647/FastLogin/pull/811/commits/d7d0caf6e7499060277b677f46e6225c0ae7b9c4, but I am not sure whether it is enough. Please, check it out.

core/src/main/java/com/github/games647/fastlogin/core/mojang/ProxyAgnosticMojangResolver.java
Enginecrafter77: Resolved in https://github.com/games647/FastLogin/pull/811/commits/edd82f7978fdb9097beaa1424ee19df5342816e9

Enginecrafter77:
> Well it's the opposite. When useProxyAgnosticResolver is enabled, it acts like the corresponding properties option is disabled as well. So we are not verifying the client IP.

Yes, I realized the typo in few minutes after I commited it, it's fixed in https://github.com/games647/FastLogin/pull/811/commits/5abd864fa5ecfe63d65ebd9f29597bfeb1a09f70.

games647 (approved):
!LGTM

Thank you

Issue #809, opened 2022-06-08
1.19 compatibility


games647:
Microsoft/Mojang changed a lot of the authentication protocol. This requires a lot of work and testing. As protocol changes are very recent, not a lot of research was done about the internal changes. Furthermore, the login authentication is a critical area, so requires careful code design and tests. As I don't have much time for a hobby project, because of personal issues, expect some delay.

# Status:
### ProtocolLib
Won't work - not tested, but strongly appears so from the code.

### ProtocolSupport
Could work - not tested

## Proxy-side
The supported proxies allows for enabling onlinemode using an API call. Therefore breaking changes are very unlikely.

### BungeeCord
Fixed in https://github.com/games647/FastLogin/commit/6665c0359b1cc45845a6b0277081f62a018ac4f9

Now always requires the latest version, because BungeeCord is backwards compatible.

### Velocity
No changes required from FastLogin.

Tander:
Can provide some info:

with ProtocolLib #559 - getting this error, autologin doesn't work
`[23:17:22 WARN]: [ProtocolLib] Loaded class com.github.games647.fastlogin.bukkit.listener.protocollib.ProtocolLibListener from FastLogin v1.11-SNAPSHOT-6665c03 which is not a depend or softdepend of this plugin.
[23:17:22 ERROR]: [FastLogin] Unhandled exception occurred in onAsyncPacket() for FastLogin
com.comphenix.protocol.reflect.FieldAccessException: No field with type com.mojang.authlib.GameProfile exists in class PacketLoginInStart.
	at com.comphenix.protocol.reflect.StructureModifier.readInternal(StructureModifier.java:216) ~[ProtocolLib.jar:5.0.0-SNAPSHOT-b559]
	at com.comphenix.protocol.reflect.StructureModifier.read(StructureModifier.java:195) ~[ProtocolLib.jar:5.0.0-SNAPSHOT-b559]
	at com.github.games647.fastlogin.bukkit.listener.protocollib.ProtocolLibListener.onLogin(ProtocolLibListener.java:126) ~[FastLoginBukkit.jar:?]
	at com.github.games647.fastlogin.bukkit.listener.protocollib.ProtocolLibListener.onPacketReceiving(ProtocolLibListener.java:96) ~[FastLoginBukkit.jar:?]
[...]

PR #821, opened 2022-06-18
1.19 Support

games647:
### Summary of your change
Add ProtocolLib (Spigot) support for 1.19

### Status

**A lot work** was necessary to put into this version and still some things are missing. However I want to inform you about the current state while I'll take a break now.

- [X] Fix username field changes in login packets for changes in 1.19
- [X] Verify client public keys against Mojang certificate
- [X] Update nonce verification to 1.19 in online mode responses from the client
- [ ] Make sending of client public keys optional
  - Depends on client implementation
  - Currently always require the key in both onlinemodes
- [ ] Forward client sent public keys to the server
- [ ] Compatibility to run on older Minecraft versions
- [ ] Allow to disable the verification of the public keys
  - Similar to vanilla configurable behavior
- [ ] Improve signature extract for better performance - see the ProtocolLib ticket

### Warning

Usage at your own caution.

### Related issue
#809 

ProtocolLib: https://github.com/dmulloy2/ProtocolLib/issues/1661


games647:
Ready for testing

Issue #786, opened 2022-04-20
Geyser Bug


JohnnyX071:
### What happened?

i tried using FastLogin with AuthMe on my crossplay server. I tested with my bedrock account and I could join. However after I relog it kept kicking me and saying: Your username contains illegal characters and for some reason said detected cracked player in the console even though I have bough bedrock account. I tried removing FaastLogin and it worked fine.

### What did you expect?

The expected behavior would be to let bedrock players always join.

### Steps to reproduce

I just used the AuthMe plugin with FastLogin

### Plugin list

Admin-GUI, AdvancedCrates, AdvancedEnchantments, AdvancedHelp, AuctionHouse, AureliumSkills, AuthMe, BKCommonLib, Citizens, ClearLag, CommandPanels, CustomCrafting, DeluxeMenus, DriveBackupV2, EliteArmor, Essentials, EssentialsChat, ExecutableItems, FastAsyncWorldEdit (WorldEdit), FastLogin, floodgate, Geyser-Spigot, HolographicDisplays, HubThat, ItemJoin, JumpPads, LuckPerms, MiningBlocks, Multiverse-Core, NexEngine, NoPlugins, PlaceholderAPI, PlugMan, ProtocolLib, Quests, SCore, ShopGUIPlus, SimpleScore, spark, SuperiorSkyblock2, SuperVanish, TAB, TelePads*, UltimateCore, UltimateCore-Farm, UltimateKits, UltraMinions, Vault, ViaBackwards, ViaVersion, VoidGen, Votifier, VotingPlugin, WolfyUtilities, WorldGuard

### Configuration file

https://gist.github.com/JohnnyX071/84d71a67eb62f7206220380b7ab4cc58

### Server log

```shell
[FastLogin] Floodgate Prefix detected on cracked player
```

### Plugin version

Beta #1069

### Platform

Spigot

### Relevance

- [X] I tried the latest build (build refers to development builds not necessary a release version)
- [X] I checked for existing tickets -
If there are, please vote them with a thumps reaction and not create new ones

Smart123s:
> #### Server log
> ```
> [FastLogin] Floodgate Prefix detected on cracked player
> ```

This message should only show up when a cracked Java player joins, and when their name contains the Floodgate prefix. Did you get this message when a bedrock player was joining?

JohnnyX071:
Yes that's when I got it. Well it doesn't matter now since I have switched to online mode. Thanks for your reply though! :)

Issue #703, opened 2022-01-17
Tried to connect as a Java Edition player!  Is Floodgate configured correctly?


rad1914:
### What happened?

When trying to auto login with FastLogin, Floodgate throws an error and disconnects some of my players

### What did you expect?

Common Auto-Login

### Steps to reproduce

I noticed that only people who log in with Bedrock and specifically a Gamertag with spaces are the ones that get logged out, the rest can log in correctly. 
I tried replacing the spaces with the "_" from the Floodgate config, but that doesn't work either

### Plugin list

AdvancedLogin, PlaceholderAPI, Oraxen, FastLogin, Spark, plugman, Worldedit, WorldGuard, EssentialsX

### Configuration file

https://gist.github.com/rad1914/1764bcca8ebe53dfa268ff075fd49b89

### Server log

```shell
https://pastebin.com/D2DgWE5S
```

### Plugin version

1.11-SNAPSHOT-3bcc6c2

### Platform

Spigot

### Relevance

- [X] I tried the latest build
- [X] I checked for existing tickets -
If there are, please vote them with a thumps reaction and not create new ones

Smart123s:
Duplicate of https://github.com/games647/FastLogin/issues/647, but no response from it's author.

> [FastLogin] Requesting premium login for registered player: Otoniel_Gamer73[m

The player was saved as a premium player, which means that auto login is needed for them (this behavior will change with https://github.com/games647/FastLogin/issues/700). This means that the player was recognized as a Floodgate player at least once. But I'm sure that they are not detected at some point, and that causes the bug.

Are you using BungeeCord/Waterfall? If not, are you using ProtocolLIb or ProtocolSupport?

fabegalo:
Bedrock players are being recognized as Java players in FastLogin causing a geyser error

fabegalo:
@games647

LuizLoyola:
Same error as him ^
Tried with a Bedrock-only and with a Bedrock+Java linked account, same error. (Java players work fine)

fabegalo:
> Same error as him ^ Tried with a Bedrock-only and with a Bedrock+Java linked account, same error. (Java players work fine)

Yes, this error is linked to Fast Login, since without Fast Login, it works normally

@games647

games647:
[...]

Issue #647, opened 2021-11-07
Bug with Floodgate (only nicknames with Space)


jassyel:
### What happened?

Players can't join if cointain spaces in nicknames

### What did you expect?

Play like someone

### Steps to reproduce
0.- Enable auto-login and auto register for floodgate in config
1.- Create a account with space (xbox)
2.- Try to join server
3.- See first time you CAN join to the server
4.- Re-join and now you can see you can't join in the server

### Plugin list

Floodgate, Geyser-Spigot, Authme, FastLogin

### Server log

```shell
https://pastebin.com/WeZmPjwM
```

### Plugin version

FastLogin version 1.11-SNAPSHOT-5612ca7

### Platform

Spigot

### Relevance

- [X] I tried the latest build
- [X] I checked for existing tickets -
If there are, please vote them with a thumps reaction and not create new ones

Smart123s:
Hi, are you using ProtocolLib or ProtocolSupport?

PR #876, opened 2022-08-09
Fix Floodgate prefix with ProtocolLib

Smart123s:
### Summary of your change
_This was originally part of https://github.com/games647/FastLogin/pull/709._
During the login start packet, the channel of the player is extracted from the ProtocolLIb handler, and if it contains a FloodgatePlayer attribute, we execute Floodgate's tasks manually, because ProtocolLib doesn't fire packets down the channel pipeline if it has an async listener attached to it.

### Related issue
Fixes https://github.com/games647/FastLogin/issues/786
Fixes https://github.com/games647/FastLogin/issues/703
Fixes https://github.com/games647/FastLogin/issues/689
Fixes https://github.com/games647/FastLogin/issues/647


games647 (approved):

games647:
Sadly this solution is incompatible with ProtocoLib v4.8.0, which is necessary for below 1.18.

#878

Smart123s:
Hopefully, I will just have to change some package/class names. Anyways, I won't be home until Sunday/Monday, so I can't check it now. If it causes too much trouble, then just revert it, or disable it for 1.18 and below with an if.

Smart123s:
I took a look at it from mobile, and playerInjectionHandler is just playerInjection in plib 4.8.0.

https://github.com/dmulloy2/ProtocolLib/blob/41bb4bacb2fb0c5a6942188dfed872b7fd4bd904/src/main/java/com/comphenix/protocol/injector/PacketFilterManager.java#L109

code above that ran successfully, and playerInjectionHandler.getChannel() exists, so this could be an easy fix, although I still can't test it until I get home. You cna check it if the issue is urgent.

Issue #885, opened 2022-08-12
java.lang.NoClassDefFoundError: com/comphenix/protocol/ProtocolLibrary when disabling


NathanFAIN:
### What happened?

When the server stops, the FastLogin plugin is disabled with an error.
It seems that ProtocolLib is disabled before FastLogin.

### What did you expect?

Disable Fastlogin without error

### Steps to reproduce

Start your server with :
![image](https://user-images.githubusercontent.com/47045209/184448718-429e6638-d9a1-4a9a-8a9f-42c616d8eba0.png)
![image](https://user-images.githubusercontent.com/47045209/184448748-def3a2e5-fd9e-46d6-bf5d-22021c6e89ab.png)
![image](https://user-images.githubusercontent.com/47045209/184448785-7cbb4358-1964-4cdb-a7b4-7df3bef3451d.png)

And stop your server.

### Plugin list

FastLogin 1.12-SNAPSHOT-feb4639, ProtocolLib 4.8.0, AuthMe 5.6.0-SNAPSHOT-b2568

### Server log

[23:28:54] [Server thread/INFO]: [ProtocolLib] Disabling ProtocolLib v4.8.0
[23:28:54] [Server thread/WARN]: [AuthMe] ProtocolLib has been disabled, unhooking packet adapters!
[23:28:54] [Server thread/INFO]: [PluginQuery] Disabling PluginQuery v1.0.36
[23:28:54] [Server thread/INFO]: [LPX] Disabling LPX v3.1.3
[23:28:54] [Server thread/INFO]: [AuthMe] Disabling AuthMe v5.6.0-SNAPSHOT-b2568
[23:28:54] [Server thread/INFO]: [AuthMe] Waiting for 0 tasks to finish
[23:28:54] [Server thread/ERROR]: [AuthMe] [STDERR] [Server thread] INFO fr.xephi.authme.libs.com.zaxxer.hikari.HikariDataSource - AuthMeMYSQLPool - Shutdown initiated...
[23:28:54] [Server thread/WARN]: Nag author(s): '[sgdc3, ljacqu, games647, Hex3l, krusic22]' of 'AuthMe' about their usage of System.out/err.print. Please use your plugin's logger instead (JavaPlugin#getLogger).
[23:28:54] [Server thread/ERROR]: [AuthMe] [STDERR] [Server thread] INFO fr.xephi.authme.libs.com.zaxxer.hikari.HikariDataSource - AuthMeMYSQLPool - Shutdown completed.
[...]

PR #892, opened 2022-08-17
Fix unregister ProtocolLib

NathanFAIN:
### Summary of your change

When FastLogin is disabled, the plugin unregisters from ProtocolLib (if ProtocolLib is present in the "plugins" folder). But it doesn't check if ProtocolLib is still loaded and not already disabled.
Accessing ProtocolLib methods when the plugin is disabled creates an error.
I modified the check to see if ProtocolLib is still loaded :

```java
if (isPluginInstalled("ProtocolLib")) { //if ProtocolLib is present in the "plugins" folder
    ProtocolLibrary.getProtocolManager().getAsynchronousManager().unregisterAsyncHandlers(this);
 }
 ```
 To
 ```java
if (getServer().getPluginManager().isPluginEnabled("ProtocolLib")) { //if ProtocolLib is still loaded
    ProtocolLibrary.getProtocolManager().getAsynchronousManager().unregisterAsyncHandlers(this);
 }
 ```

### Related issue

Fixes/Related #885

### DIFF
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/FastLoginBukkit.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/FastLoginBukkit.java
@@ -187,7 +187,7 @@ public class FastLoginBukkit extends JavaPlugin implements PlatformPlugin<Comman
             }
         }
 
-        if (isPluginInstalled("ProtocolLib")) {
+        if (getServer().getPluginManager().isPluginEnabled("ProtocolLib")) {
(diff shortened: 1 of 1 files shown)


games647:
> if ProtocolLib is still loaded and not already disabled.

Shouldn't plugins be disabled in the order they were enabled? The code was added, because Paper complained about an asynchronous task still running by FastLogin on shutdown.

games647 (approved):
Okay still works

!LGTM

Thank you

Issue #1199, opened 2024-05-04
java.lang.NoClassDefFoundError: fastlogin/gson/GsonBuilder due to update dependency


Freyhold:
### What happened?

In last commit 8e35c5d @games647 removed 
```
        <dependency>
            <groupId>com.google.code.gson</groupId>
            <artifactId>gson</artifactId>
            <version>2.10.1</version>
        </dependency>
```
after add this dependency to core, plugin works.  To solve this i made pull request #1200

### Server log

[19:49:12 ERROR]: Error occurred while enabling FastLogin v1.12-SNAPSHOT-d58ef6b (Is it up to date?)
java.lang.NoClassDefFoundError: fastlogin/gson/GsonBuilder
        at com.github.games647.craftapi.resolver.AbstractResolver.<init>(AbstractResolver.java:48) ~[FastLoginBukkit (1).jar:?]
        at com.github.games647.craftapi.resolver.MojangResolver.<init>(MojangResolver.java:43) ~[FastLoginBukkit (1).jar:?]
        at com.github.games647.fastlogin.core.shared.FastLoginCore.load(FastLoginCore.java:122) ~[FastLoginBukkit (1).jar:?]
        at com.github.games647.fastlogin.bukkit.FastLoginBukkit.onEnable(FastLoginBukkit.java:89) ~[FastLoginBukkit (1).jar:?]
        at org.bukkit.plugin.java.JavaPlugin.setEnabled(JavaPlugin.java:287) ~[paper-api-1.20.4-R0.1-SNAPSHOT.jar:?]
        at io.papermc.paper.plugin.manager.PaperPluginInstanceManager.enablePlugin(PaperPluginInstanceManager.java:188) ~[paper-1.20.4.jar:git-Paper-493]
        at io.papermc.paper.plugin.manager.PaperPluginManagerImpl.enablePlugin(PaperPluginManagerImpl.java:104) ~[paper-1.20.4.jar:git-Paper-493]
        at org.bukkit.plugin.SimplePluginManager.enablePlugin(SimplePluginManager.java:507) ~[paper-api-1.20.4-R0.1-SNAPSHOT.jar:?]
        at org.bukkit.craftbukkit.v1_20_R3.CraftServer.enablePlugin(CraftServer.java:639) ~[paper-1.20.4.jar:git-Paper-493]
[...]

PR #1200, opened 2024-05-04
add removed dependency in last commit 8e35c5d that updated dependency

Freyhold:
### Summary of your change
add missing dependency that was removed by games647 

### Related issue
Issue #1199


games647 (approved):

games647:
Thanks, I didn't noticed that.

Issue #1222, opened 2024-07-06
Plugin does not function correctly at all using Velocity


Clyde6790pGIT:
### What happened?

Floodgate and Premium users get auto registered but sometimes it does not at all, and sometimes when it does auto register it doesn't auto log you in, trapping users in an "invalid password" loop. 
@Smart123s 

### Steps to reproduce

1. Just log into the server as a Floodgate or Premium user
2. Get through the insanely buggy autoregister
3. Rejoin, and wow your locked out.
![image](https://github.com/games647/FastLogin/assets/78290310/4f35650f-7446-450a-bf89-dc58d7aca7ca)

### Plugin list

LoginSecurity, Fastlogin, GeyserMC, Floodgate, GeyserUpdater, LimboAPI, LimboReconnect, SkinsRestorer, ViaVersion/Backwards, VelociTab, OpenAudioMC

### Configuration file

https://hastebin.com/share/naxoduheti.yaml

### Server log

https://hastebin.com/share/atayigadem.csharp
(IP's are just localhost so they aren't redacted)

### Plugin version

1.12-SNAPSHOT-29100b5

### Platform

Velocity

### Relevance

- [X] I tried the [latest build](https://ci.codemc.io/job/Games647/job/FastLogin/) 
(build refers to development builds not necessary a release version; i.e. v1.10 is out of date)

- [X] I checked for existing tickets -
If there are, please vote them with a thumbs reaction and not create new ones

Clyde6790pGIT:
Hi, I tested it again and It doesn't even WORK PROPERLY, it doesn't even make it to auto registration anymore nor even auto login, premium or floodgate players now get treated as cracked players and I almost want to give up on this plugin, it worked fine for a bit before it just stopped working.

games647:
Please post the corresponding spigot log, because that's then responsible to execute the force login command.

Clyde6790pGIT:
Hi, here are logs from PaperMC.
https://hastebin.com/share/ijasahaloc.csharp

Clyde6790pGIT:
I've also noticed a weird pattern, it never works on the first server setup in the Velocity server list, but works on the 2nd and 3rd one, but if you swap the 2nd or 3rd one into the first one, it doesn't work.

Smart123s:
Is your config form Velocity? If so, how do you use SQLite? If not, could you share your FastLoginVelocity config? Most of the stuff in FastLoginBukit config is ignored when using a proxy.

[...]

Issue #1176, opened 2024-03-27
Velocity auto register and login doesn't work for floodgate players 


bobhenl:
### What happened?

Hello, I thought this issue was fixed for velocity in the past but it probably occurred again. With the latest velocity, fast login Dev build, floodgate and geyser last builds, geyser players aren't automatically registered and logged in.

Just a note it work on waterfall I had to switch there, but it would be great if it can work on velocity too

There are already many issues about it like https://github.com/games647/FastLogin/issues/729#issuecomment-1050530340

I even tried the build from here that was supposed to work but it isn't anymore 

Please test it with the latest releases and you'll see that it doesn't work anymore 

### What did you expect?

That geyser floodgate players would be automatically registered and logged in the same way as the java players are and the same way as it works with waterfall now 

### Plugin version

1298 the latest build currently 

### Platform

Velocity

### Relevance

- [X] I tried the [latest build](https://ci.codemc.io/job/Games647/job/FastLogin/) 
(build refers to development builds not necessary a release version; i.e. v1.10 is out of date)

- [X] I checked for existing tickets -
If there are, please vote them with a thumbs reaction and not create new ones

Issue #1070, opened 2023-08-07
Auto login for floodgate is not working.


hasankayra04:
### What happened?

Fastlogin detects it is a floodgate player but doesn't auto login.

### Plugin list

HamsterAPI, ModChecker, Skript, PlaceholderAPI, DisableJoinMessage, ExploitFixer, MOTD, ViaVersion, floodgate, MyCommand, Floodgate-Skript, TitleManager, ClientDetector, SkinsRestorer, FastLogin, AuthMe, HolographicDisplays, ProtocolLib, Log4JExploitFix, CasualProtector

### Configuration file

https://hastebin.com/share/afobitawuj.yaml

### Server log

https://mclo.gs/GTkSGZy

### Plugin version

1.12-SNAPSHOT-${git.commit.id.abbrev}

### Platform

BungeeCord

### Relevance

- [X] I tried the latest build (build refers to development builds not necessary a release version)
- [X] I checked for existing tickets -
If there are, please vote them with a thumps reaction and not create new ones

hasankayra04:
I noticed server logs a incorrect command while floodgate player logges in. Maybe fastlogin tries to login but fails due to command being wront/incomplete?

hasankayra04:
Edit: Yeah incorrect command comes from one of my skripts. Unrelated.

HierarchyMan:
I really need this to be fixed
im experiencing the same issue

Issue #991, opened 2023-02-23
can't work with floodgate on velocity


BingYanchi:
### What happened?

I used the floodgate link to create a genuine Minecraft account, but an error occurred when I logged in.
Use latest Velocity, Floodgate, FastLogin.

### What did you expect?

normal access server

### Steps to reproduce

- Use floodgate link genuine account
- Login to the server on Velocity
- Can't enter, error occurred

### Plugin list

Floodgate, FastLogin

### Configuration file

https://gist.github.com/BingYanchi/cd344caca2dccc0911cb06d1552f3448

### Server log

https://gist.github.com/BingYanchi/37a3877651b3e581844db495a6a89dac

### Plugin version

Build #1195

### Platform

Velocity

### Relevance

- [X] I tried the latest build (build refers to development builds not necessary a release version)
- [X] I checked for existing tickets -
If there are, please vote them with a thumps reaction and not create new ones

Smart123s:
FastLogin Velocity doesn't support Floodgate yet. Duplicate of https://github.com/games647/FastLogin/issues/729#issuecomment-1050530340

PR #1223, opened 2024-07-06
Floodgate Velocity fixes

Smart123s:
### Summary of your change

First of all, I'm sorry for this awful reflection heavy code. I didn't want to add too many dependencies, so there are some type safety related warnings.

Two commits in this PR, but the second one is a one liner, and I'm lazy.

9a40cf0afb1d59d77892b098ca8fdbc13d33b4b9: After running the LOGIN event, it didn't get marked as run due to an early return. In Velocity, this caused all proxy commands to have no effect.

2d177e3df563fe30b4f8ab4ec42c07a8761fda07: Most of the Floodgate checks are run at the PreLoginEvent, which requires precise usernames. However, Floodgate only sets the correct username in GameProfileRequestEvent ([see here](https://github.com/GeyserMC/Floodgate/blob/49bd56446fa10931b1c4abf6609055571a4ae105/velocity/src/main/java/org/geysermc/floodgate/listener/VelocityListener.java#L185)). If username prefixes are used, this made FastLogin unable to detect Floodgate players at this stage of the login process.

### Related issue

Should fix pretty much every Floodgate & Velocity related issue.
Fixes https://github.com/games647/FastLogin/issues/1222
Fixes https://github.com/games647/FastLogin/issues/1176
Fixes https://github.com/games647/FastLogin/issues/1070
Fixes https://github.com/games647/FastLogin/issues/991
Fixes https://github.com/games647/FastLogin/issues/935
Fixes https://github.com/games647/FastLogin/issues/729

### DIFF
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/ConnectionListener.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/ConnectionListener.java
@@ -90,6 +90,7 @@ public class ConnectionListener implements Listener {
                 if (floodgatePlayer != null) {
                     Runnable floodgateAuthTask = new FloodgateAuthTask(plugin.getCore(), player, floodgatePlayer);
                     Bukkit.getScheduler().runTaskAsynchronously(plugin, floodgateAuthTask);
+                    plugin.getBungeeManager().markJoinEventFired(player);
                     return;
                 }
             }
--- a/velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
+++ b/velocity/src/main/java/com/github/games647/fastlogin/velocity/FastLoginVelocity.java
@@ -230,4 +230,8 @@ public class FastLoginVelocity implements PlatformPlugin<CommandSource> {
     public UUID getProxyId() {
         return proxyId;
     }
+
+    public ProxyServer getServer() {
+        return server;
+    }
 }
--- a/velocity/src/main/java/com/github/games647/fastlogin/velocity/listener/ConnectListener.java
+++ b/velocity/src/main/java/com/github/games647/fastlogin/velocity/listener/ConnectListener.java
@@ -38,6 +38,7 @@ import com.github.games647.fastlogin.velocity.task.FloodgateAuthTask;
 import com.github.games647.fastlogin.velocity.task.ForceLoginTask;
 import com.google.common.cache.Cache;
 import com.google.common.collect.ListMultimap;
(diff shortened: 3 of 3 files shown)


Clyde6790pGIT:
Hi, I am testing this and it worked almost instantly, it did tell me to register a few times but then it registered automatically and logged me in, this also worked with authmevelocity.

games647:
Oh wow, that's really nasty reflection. Sadly we cannot use the other `ProfileEvent`, because the decision needs to made earlier. I guess we cannot use the API at that instant?

Nevertheless, thank you very much for the research.

Smart123s:
> I guess we cannot use the API at that instant?

Both the name and the UUID are set at the GameProfile stage.

The only thing we have access to at this step is the username, and an offline UUID.
The purpose of username prefixes is that a Floodgate user and a Java user can be online at the same time, so that's not really useful if we don't know which platform the player is on.
As for offline UUID, AFAIK, Minecraft generates them based on usernames, so same problem there.

> Sadly we cannot use the other ProfileEvent, because the decision needs to made earlier.

Yeah, I know. The Minecraft login procedure is pretty strict.

A more prettier solution would be to change Floodgates code to directly expose that cache variable, but they have [previously stated](https://github.com/GeyserMC/Geyser/issues/1316#issuecomment-695788722) that they don't want to support offline-mode servers.

games647 (approved):

Tim203:
I was just checking some Floodgate related issues/PRs to see if https://github.com/GeyserMC/Floodgate/issues/143 is still an issue, and I saw this PR. This indeed looks like a lot of reflection and looks very prune to breaking. Instead, on all platforms we support, we have a channel attribute key called `floodgate-player` that returns `FloodgatePlayer`. This will probably also break in Floodgate 3.0 (because we'll be combining the Floodgate API with the Geyser API) as it'll return `Connection` (a [base api](https://github.com/GeyserMC/api) class) instead, but that seems much more future proof.

Smart123s:
@Tim203 you are absolutely right that this is not an optimal implementation.
However, Velocity doesn't expose the Channel so I had to use Reflections either way. ([see Floodgate's code](https://github.com/GeyserMC/Floodgate/blob/06f2ed966ed82ea20b70757137d84a904ce177a8/velocity/src/main/java/org/geysermc/floodgate/listener/VelocityListener.java#L121-L140))
In that situation, it didn't really matter whether I use Reflections to extract the Channel from Veloctiy, or use Reflections to extract the FloodgatePlayer from Floodgate.

If Floodgate API 3.0 drops, I might switch to extracting the Channel from Velocity.

PR #1269, opened 2024-12-01
Fix ProtocolLib version and paper URL

Elikill58:
### Summary of your change

- Use new URL for papermc. The old one will breaked at the end of december
- The used version of ProtocolLib is no longer available. I changed it to new one.


games647:
Thanks a lot for the heads up.

games647 (approved):

PR #1283, opened 2025-03-28
Added `/fldelete <name>` to delete profiles from storage

tbm00:
Was running into errors after converting players premium<->cracked that required deleting the players' entries from SQL to fix.

### DIFF
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/command/DeleteCommand.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/command/DeleteCommand.java
@@ -1,0 +1,85 @@
+/*
+ * SPDX-License-Identifier: MIT
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015-2024 games647 and contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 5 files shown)


bukkit/src/main/resources/plugin.yml
```diff
@@ -62,3 +67,7 @@ permissions:
         description: 'Label others as cracked'
         children:
             ${project.artifactId}.command.cracked: true
+
+    ${project.artifactId}.command.delete:
+        description: 'Delete other players profile data'
+        default: true
```
TuxCoding, line 73: I guess this shouldn't be true right ;)

A normal player shouldn't be able to delete data.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/command/DeleteCommand.java
```diff
@@ -0,0 +1,85 @@
...
+            plugin.getCore().sendLocaleMessage("no-permission", sender);
+            return true;
+        }
+
+        if (args.length < 1) {
+            sender.sendMessage("Error: Must supply username to delete!");
+            return false;
+        }
+
+        int count = plugin.getCore().getStorage().deleteProfile(args[0]);
```
TuxCoding, line 63: This executes synchronously. It should be asynchronously, because it uses network blocking I/O.

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/command/DeleteCommand.java
```diff
@@ -0,0 +1,85 @@
...
+import org.bukkit.Bukkit;
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.command.ConsoleCommandSender;
+import org.bukkit.command.TabExecutor;
+import org.bukkit.entity.Player;
+
+import com.github.games647.fastlogin.bukkit.FastLoginBukkit;
+
+public class DeleteCommand implements TabExecutor {
```
TuxCoding, line 40: This will only work if you use it without BungeeCord. At least add some safeguard if BungeeCord is in use.

TuxCoding (changes requested):

bukkit/src/main/java/com/github/games647/fastlogin/bukkit/command/DeleteCommand.java
tbm00: I've never worked with BungeeCord before, so I'm not sure how it should be properly be implemented.

But I did add the safeguard!

PR #1304, opened 2025-06-08
Remap User UUID Functionality - Idea?

WuzzyLV:
### Summary of your change

I'm not sure if im using the wrong settings or so on, but ive always had issues with players changing names and having to manually roll everything over, one day i noticed that it was overriding the UUIDs anyways (premiumUuid: false), so i made this.

Is this a valid QoL thing or am i just using the plugin wrong?

_Code for explanation sake, not expecting to be merged_


TuxCoding:
The issue is that the UUID is only available after authentication. Therefore, you can only rely on the username and the existing data of the auth plugin. We can only do this name change migration then for onlinemode connections. 

Furthermore, most auth plugins only rely on the username which complicates it even further, because they have no way to migrate the data. Only way would be to create a separate account, but this would use a different password, because FastLogin doesn't save it.

PR #1307, opened 2025-06-13
Fix Passky Hook and implement debug logging to Passky

zigazajc007:
### Summary of your change

Fix Passky Hook and implement debug logging to Passky


TuxCoding:
Merged manually, because I think debug logging is disabled. So I changed it to error level.

PR #1365, opened 2026-07-20
Fix incompatibility by skipping offine encrypt instead of kick

Lumine1909:

### Summary of your change
Do not kick player when processing encryption on offline players, so the later netty handler can work correctly.

[//]: # (Example: motivation, enhancement)

### Motivation
I'm currently maintaining [a plugin that enable encryption on offline players](https://github.com/Lumine1909/OfflineEncryptor), and it is incompatible with FastLogin on both online and offline players. I added a fix  in my plugin that resolved online player encrypted twice. But currently FastLogin kick the offline player due to exception when processing encryption related packet, since their `verifyToken` is a empty array.

I added a check for offline player when doing encryption, so that my plugin can process the encrypt response packet later : )

### Related pr
https://github.com/Lumine1909/OfflineEncryptor/pull/6

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved login handling for offline players by skipping unnecessary encryption-token verification.
  - Preserved existing behavior for invalid login sessions, including session rejection.

### DIFF
--- a/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/ProtocolLibListener.java
+++ b/bukkit/src/main/java/com/github/games647/fastlogin/bukkit/listener/protocollib/ProtocolLibListener.java
@@ -178,7 +178,7 @@ public class ProtocolLibListener extends PacketAdapter {
         if (session == null) {
             plugin.getLog().warn("Profile {} tried to send encryption response at invalid state", sender.getAddress());
             sender.kickPlayer(plugin.getCore().getMessage("invalid-request"));
-        } else {
+        } else if (session.getVerifyToken().length != 0) { // Skip offline player
             byte[] expectedVerifyToken = session.getVerifyToken();
             if (verifyNonce(sender, packetEvent.getPacket(), session.getClientPublicKey(), expectedVerifyToken)) {
                 packetEvent.getAsyncMarker().incrementProcessingDelay();


TuxCoding:
Hi @Lumine1909 Thanks for the suggestion. It would need a little bit more documentation to be included in this project. First of all what does the client if it is a offline player which we requested encryption from and second does your plugin make any modifications of FastLogin state? If the latter is the case, this behavior should be documented or better we might need to add an event, so it's much more future proof.

Lumine1909:
> Hi @Lumine1909 Thanks for the suggestion. It would need a little bit more documentation to be included in this project. First of all what does the client if it is a offline player which we requested encryption from and second does your plugin make any modifications of FastLogin state? If the latter is the case, this behavior should be documented or better we might need to add an event, so it's much more future proof.

Thank you for the reply! For your question:

1. FL does not process offline player and pass its C2SHello to the following handler. Then OE will [intercept this packet and start encryption protocol](https://github.com/Lumine1909/OfflineEncryptor/blob/main/paper/src/main/java/io/github/lumine1909/offlineencryptor/paper/PaperPacketInterceptor.java). And when C2SKey arrived, currently FL will throw an exception since that player is not registered as a online user. Since OE handles everything afterward, FL can pass it through instead of throw, as what I modified in this pr. 

2. OE doesn't modify any states of FL on bukkit, since its handler is down the pipeline, it [only check if FL has finished the online process](https://github.com/Lumine1909/OfflineEncryptor/blob/main/compatibility/src/main/java/io/github/lumine1909/offlineencryptor/compat/AuthenticateCompats.java), it works well rn but is fragile if FL changes its relevant logic in the future. However, for Velocity, since FL implement the feature by listen to VC's event that triggers later than netty handler, the approach in OE is quite hacky. But it also doesn't touch FL's internal state.

